PR #49, opened 2018-01-16
Transition to AST v3

Yoric:


src/source/mod.rs
```diff
@@ -5,19 +5,6 @@
...
-/// Using Node.js + Esprima to load an AST.
-pub mod esprima;
-#[cfg(feature = "esprima")]
-pub use self::esprima::Esprima;
-
-/// Using Node.js + Babel to load an AST.
-pub mod babel;
-pub use self::babel::Babel;
\ No newline at end of file
+/// Parsing JavaScript using the Shift source parser (in Node).
```
xtuc: I'm wondering why do you use Shift instead of Babel now? Did you encounter an issue with Babel?

Yoric: We didn't encounter any problem with the parser. However, it turns out that the very permissive ESTree-style structure of the Babel AST makes it pretty hard (read "slow") to check whether an AST is correct. So we decided to transition to a Shift-style AST. Early experiments indicate that it's much easier/faster to check correctness.

Note that we're slightly patching the AST after the Shift parser to make it even easier/faster to check.

PR #63, opened 2018-03-16
fix shift-parser lib missing error

sendilkumarn:


Yoric:
Thanks!

Note that we can actually get rid of babel/babylon, since we have switched to the shift parser.

PR #69, opened 2018-03-16
Increase memory used by node

### LINKED ISSUE #67: Make it possible to parse concat.js
Currently, concat.js is too large for the Shift parser. We should find a way to change the options of Node to avoid running out of heap space.

sendilkumarn:
fixes #67 
This adds `--max_old_space_size=4096` to the node


src/source/babel.rs
```diff
@@ -61,6 +61,7 @@ impl Babel {
 
         debug!(target: "Babel", "Launching script {}", script);
         let output = Command::new(&*self.bin_path)
+            .arg("--max_old_space_size=4096")
```
Yoric: Ah, I should remove this module entirely.

src/source/esprima.rs
```diff
@@ -88,6 +88,7 @@ impl SourceParser for Esprima {
 
         info!(target: "Esprima", "Reading path {}", path);
         let output = Command::new(&*self.bin_path)
+            .arg("--max_old_space_size=4096")
```
Yoric: I should remove that module, too :)

src/source/spidermonkey.rs
```diff
@@ -36,6 +36,7 @@ impl SpiderMonkey {
     fn parse_script_output(&self, script: &str) -> Result<JSON, Error> {
         info!(target: "SpiderMonkey", "Launching script {}", script);
         let output = Command::new(&*self.bin_path)
+            .arg("--max_old_space_size=4096")
```
Yoric: That one doesn't make sense, since it doesn't launch Node. Still, I should remove the module anyway :)

src/source/shift.rs
```diff
@@ -65,6 +65,7 @@ impl Shift {
 
         debug!(target: "Shift", "Launching script {}", script);
         let mut child = Command::new(&*self.bin_path)
+            .arg("--max_old_space_size=4096")
```
Yoric: Ok, this one is the interesting module.

Ideally, we should be able to customize this value without changing the code, but this can wait for a followup patch.

Could you just add a comment explaining why we're doing this?

src/bin/encode.rs
```diff
@@ -88,7 +88,7 @@ fn handle_path<'a>(options: &Options<'a>,
         .expect("Could not open source")
         .len();
 
-    println!("Parsing.");
+    println!("Parsingq.");
```
Yoric: Please fix this typo :)

src/source/babel.rs
```diff
@@ -157,6 +158,7 @@ impl SourceParser for Babel {
 
     /// Parse a text source file, using Babel.
     fn parse_file<P: AsRef<Path>>(&self, path: P) -> Result<JSON, Error> {
+        println!("babel");
```
Yoric: Please remove this `println`.

src/source/esprima.rs
```diff
@@ -83,11 +84,13 @@ impl SourceParser for Esprima {
 
     /// Parse a text source file, using Esprima.
     fn parse_file<P: AsRef<Path>>(&self, path: P) -> Result<JSON, Error> {
+        println!("prima");
```
Yoric: Please remove this `println`.

src/source/shift.rs
```diff
@@ -160,6 +161,7 @@ impl SourceParser for Shift {
 
     /// Parse a text source file, using Shift.
     fn parse_file<P: AsRef<Path>>(&self, path: P) -> Result<JSON, Error> {
+        println!("shift");
```
Yoric: Please remove this `println`.

I assume that there are still some `println!` in the code but if you wish to add logging, you should use macros `debug!`, `info!` or `warn!` and set a target, so that we can explicitly enable/disable them.

See https://docs.rs/log/0.4.1/log/ for the documentation on these macros.

src/source/spidermonkey.rs
```diff
@@ -80,6 +81,7 @@ impl SourceParser for SpiderMonkey {
 
     /// Parse a text source file, using SpiderMonkey.
     fn parse_file<P: AsRef<Path>>(&self, path: P) -> Result<JSON, Error> {
+        println!("spiderMonkey");
```
Yoric: Again, please remove the `println`.

src/bin/encode.rs
sendilkumarn: err... this was in my old commit 😁 Let me rebase this :)

src/source/shift.rs
sendilkumarn: I have done this for debugging.

Yoric: Sure. Just don't forget to remove them before you submit a patch :)

Yoric:
Sorry, I merged this by accident.
Can you add the test I requested before I actually land the patch?

Yoric:
If the test I requested is not possible, can you just add the samples from `facebook/single/**/*js`?

PR #70, opened 2018-03-16
Add zero and one literal

### LINKED ISSUE #28: Investigate having ZeroLiteral and OneLiteral
Popular wisdom tells us that the most common numbers are `0` and `1`. Instead of taking 1 varnum + 8 bytes for these numbers, we could reduce them to 1 varnum each, at the expense of increasing the number of entries and the length of some varnums.

sendilkumarn:
fixes #28 

check [here](https://github.com/binast/binjs-ref/issues/28#issuecomment-373871583) for detailed analysis

### DIFF
--- a/src/source/shift.rs
+++ b/src/source/shift.rs
@@ -296,6 +296,20 @@ impl FromShift {
                 // - `remove` is the `VariableDeclarationStatement`
                 // - `object` is the `VariableDeclaration`
             }
+            Some("LiteralNumericExpression") => {
+                let remove = match object.remove("value") {
+                    Some(JSON::Number(remove)) => remove,
+                    _ => panic!("No field `declaration` in a `VariableDeclarationStatement`")
+                };
+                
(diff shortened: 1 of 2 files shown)


sendilkumarn:
@Yoric We do have 3% change, WDYT about this ?

src/source/shift.rs
```diff
@@ -296,6 +296,20 @@ impl FromShift {
                 // - `remove` is the `VariableDeclarationStatement`
                 // - `object` is the `VariableDeclaration`
             }
+            Some("LiteralNumericExpression") => {
+                let remove = match object.remove("value") {
```
Yoric: That code is a bit complicated, but good enough for an experiment :)

Yoric:
Ok, this looks conclusive. Not landable, but definitely worth investigating!

Filed followup as https://github.com/binast/binjs-ref/issues/77.

PR #75, opened 2018-03-17
removed deprecated parsers

### LINKED ISSUE #72: Remove deprecated parsers
We have modules for parsing with Esprima, SpiderMonkey, Babel. These modules are not maintained and should be removed.

sendilkumarn:
fixes #72


Yoric:
Thanks!

PR #89, opened 2018-03-26
Pass NODE_MAX_OLD_SPACE_SIZE as an env variable

sendilkumarn:


examples/compare_compression.rs
```diff
@@ -64,6 +64,10 @@ fn main() {
                 .takes_value(true)
                 .possible_values(&["identity", "gzip", "br", "deflate"])
                 .help("Compression format for the binjs files"),
+            Arg::with_name("nodememory")
+                .long("BINAST_NODE_MAX_OLD_SPACE_SIZE")
```
Yoric: That name is a bit long :)

sendilkumarn: yeah. Let me 🤔 over it

Yoric: Also, shouldn't you require it to be a number?

sendilkumarn: It doesnt matter, anyway we have to typecast it later right ?

Yoric: Yes, but casting it earlier makes for better error messages/documentation.

examples/compare_compression.rs
```diff
@@ -64,6 +64,10 @@ fn main() {
...
+                .takes_value(true)
+                .help("To increase memory for running node process"),
```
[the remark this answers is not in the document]
sendilkumarn: I think it is from V8

examples/compare_compression.rs
Yoric: It's not very nice, but you can add something along the lines of

```rust
    .validator(|x| match x.parse<u32>() {
        Ok(_) => Ok(()),
        Err(e) => Err(format!("Invalid argument {:?}", e))
    })
```

to ensure that it's a proper integer.

examples/compare_compression.rs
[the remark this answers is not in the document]
Yoric: Sure, but you have to define it if you use it :) At the very least, mention that it's equivalent to arg foo and foo of Node.

examples/compare_compression.rs
sendilkumarn: It is kinda repetitive, you have to write this thing everywhere, if we use validator 😞

Yoric: Maybe we could somehow make it part of `shift.rs`?

Yoric: Ok, you have convinced me: let's switch to plan B and make it an environment variable. This will be less invasive :)

sendilkumarn: can we validate it inside shift ?

sendilkumarn: ha ha, I didnt refresh. 

Environment variable 🎉 
But I have no idea about it. Let me explore

Yoric: 1/ Perhaps Node already provides an Environment variable for that purpose? If so, we just need to document it.

2/ Otherwise, we can use an environment variable `BINAST_NODE_MAX_OLD_SPACE_SIZE` and use `std::env` to decode it in the `Shift` module.

src/bin/encode.rs
```diff
@@ -188,6 +188,9 @@ fn main() {
             Arg::with_name("statistics")
                 .long("show-stats")
                 .help("Show statistics."),
+            Arg::with_name("dump")
```
Yoric: Please don't undo that change.

Yoric: (we'll need to readd the `dump` option, but first we need to reimplement it :) )

src/source/shift.rs
```diff
@@ -64,7 +65,18 @@ impl Shift {
         script);
 
         debug!(target: "Shift", "Launching script {}", script);
+
+        let node_memory = match env::var("MAX_OLD_SPACE_SIZE") {
```
Yoric: I'd prefer if we called it `NODE_MAX_OLD_SPACE_SIZE`.

src/source/shift.rs
```diff
@@ -64,7 +65,18 @@ impl Shift {
...
+            Err(_) => String::from("--max_old_space_size=2048"),
+            Ok(v) => {
+                let mut node_arg = String::from("--max_old_space_size=");
```
Yoric: The following is more idiomatic:

```rust
format!("--max_old_space_size={}", v)`
```

sendilkumarn:
@Yoric updated  :)

README.md
```diff
@@ -39,6 +39,7 @@ git clone https://github.com/binast/binjs-ref
 cargo run --bin binjs_encode -- --help
 cargo run --bin binjs_decode -- --help
 ```
+Note: If you need to increase the memory for the node process by setting the environment variable `NODE_MAX_OLD_SPACE_SIZE=2048` (memory in MB, similar to node's `--max_old_space_size`). By default it uses 2048MB.
```
Yoric: I'd rephrase

```markdown
**Note** The JS parser may choke on very large JS source files. If so, you'll need to set the environment variable `NODE_MAX_OLD_SPACE_SIZE=xxxx`. This will instruct the Node-based parser to allocate more memory. The default value is 2048 (Mb). This is equivalent to passing `--max_old_space_size` to the Node process.
```

Yoric:
Also, don't forget to rebase and change the name of the patch :)

Yoric:
Looks good to me. Flatten the changes and I'll merge them!

sendilkumarn:
Done :)

Yoric:
...and merge!

Thanks for the patch :)

PR #95, opened 2018-04-09
Add missing node name into multipart grammar document

arai-a:
In multipart document, "node name" seems to be missing in the grammar table, compared to the implementation,
so added it to the document.

also, I cannot find the implementation for reading fields.
I haven't touched that document tho.


Yoric:
Thanks!

PR #97, opened 2018-04-14
Issue 91 - Resolve stack overflows

Yoric:
This PR moves some of the work from the main thread to dedicated threads in test_roundtrip and binjs_encode.

For the moment, we have one thread per task (there are a total of three tasks), but we should probably move the entire work to a dedicated thread, this would be simpler.


src/bin/encode.rs
```diff
@@ -150,7 +149,20 @@ fn handle_path<'a>(options: &Options<'a>,
 
     println!("Successfully compressed {} bytes => {} bytes", source_len, dest_len);
 }
+
 fn main() {
+    thread::Builder::new()
+        .name("large stack dedicated thread".to_string())
+        .stack_size(20 * 1024 * 1024)
```
sendilkumarn: interesting 🤔

PR #116, opened 2018-05-05
Don't launder away is_skippable when deanonymizing types

efaust:
Looks like we just forgot to forward this information when doing type deanonymization.

Is this supposed to bump some micro-version?


Yoric:
We probably want to release a new version of this module soon anyway.

PR #127, opened 2018-05-11
Nested scope closure

efaust:
There are two bugs here:

First, catch blocks had mismatched stack pushes and pops, leading to total confusion.
Second, we were blurring all seen names into flat levels, meaning that we couldn't tell sibling uses and bindings from nested uses and bindings.

Previously,

```
var a;
(function () {
    a;
    {
        let a;
    }
})
```

did not mark the var as closed over, even though it clearly is, since it saw the sibling binding.

### DIFF
--- a/crates/binjs_es6/src/scopes.rs
+++ b/crates/binjs_es6/src/scopes.rs
@@ -113,12 +113,15 @@ impl AnnotationVisitor {
         }
     }
 
-    fn push_var_scope(&mut self, path: &Path) {
-        debug!(target: "annotating", "push_var_scope at {:?}", path);
+    fn push_incomplete_var_scope(&mut self, _path: &Path) {
         self.var_names_stack.push(HashSet::new());
         self.lex_names_stack.push(HashSet::new());
-        self.push_free_names();
+    }
+    fn push_var_scope(&mut self, path: &Path) {
+        debug!(target: "annotating", "push_var_scope at {:?}", path);
+        self.push_incomplete_var_scope(path);
         self.push_direct_eval();
+        self.push_free_names();
     }
     fn pop_incomplete_var_scope(&mut self, path: &Path) -> VarAndLexNames {
         debug!(target: "annotating", "pop_incomplete_var_scope at {:?}", path);
@@ -304,7 +307,7 @@ impl Visitor<()> for AnnotationVisitor {
         // so we introduce a var scope in `catch(ex)`, as if `catch(ex) { ... }` was
         // a function.
 
-        self.push_var_scope(path);
+        self.push_incomplete_var_scope(path);
         self.push_param_scope(path);
         Ok(VisitMe::HoldThis(()))
     }
--- a/crates/binjs_es6/src/scopes.rs
+++ b/crates/binjs_es6/src/scopes.rs
@@ -1,7 +1,7 @@
 use ast::*;
 use binjs_shared::{ FromJSON, ToJSON, VisitMe };
 
-use std::collections::{  HashSet };
+use std::collections::{  HashSet, HashMap };
 
 use itertools::Itertools;
 use json::JsonValue as JSON;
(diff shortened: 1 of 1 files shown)


crates/binjs_es6/src/scopes.rs
```diff
@@ -55,16 +53,23 @@ impl AnnotationVisitor {
...
-        for name in free_names_in_current_function.drain() {
+    fn pop_free_names(&mut self, bindings: &[&HashSet<String>], function_scope: bool) {
+        let mut free_names_in_current_block = self.free_names_in_block_stack.pop().unwrap();
+        for (name, old_cross_function) in free_names_in_current_block.drain() {
             let is_bound = bindings.iter()
                 .find(|container| container.contains(&name))
                 .is_some();
             if !is_bound {
-                self.free_names_in_nested_functions.insert(name);
+                let parent_free = self.free_names_in_block_stack.last_mut();
```
Yoric: Nit:
```rust
if let Some(mut parent_free) = self.free_names-in_block_stack.last_mut() {
  // ...
}
```

Yoric: Nit:
```rust
if let Some(mut parent_free) = self.free_names-in_block_stack.last_mut() {
  // ...
}
```

and get rid of `unwrapped`.

crates/binjs_es6/src/scopes.rs
```diff
@@ -55,16 +53,23 @@ impl AnnotationVisitor {
...
+                if parent_free.is_some() {
+                    let mut unwrapped = parent_free.unwrap();
+                    if unwrapped.get(&name) == Some(&true) {
```
Yoric: Nit: Here, you should inspect a value, rather than create a new one for comparison purposes.

```rust
if let Some(&true) = unwrapped.get(&name) {
  // ...
}
```

Except in this case, it looks like you want to update the entry, so it's rather something along the lines of

```rust
let new_is_cross_function = old_cross_function || function_scope;

unwrapped.entry(&name)
  // Modify the entry if it is already present.
  .and_modify(new_is_cross_function)
  // Or insert a new value.
  .or_insert(new_is_cross_function);
```

crates/binjs_es6/src/scopes.rs
```diff
@@ -55,16 +53,23 @@ impl AnnotationVisitor {
     }
 
     fn push_free_names(&mut self) {
-        self.free_names_in_function_stack.push(HashSet::new());
+        self.free_names_in_block_stack.push(HashMap::new());
     }
-    fn pop_free_names(&mut self, bindings: &[&HashSet<String>]) {
-        let mut free_names_in_current_function = self.free_names_in_function_stack.pop().unwrap();
-        for name in free_names_in_current_function.drain() {
+    fn pop_free_names(&mut self, bindings: &[&HashSet<String>], function_scope: bool) {
```
Yoric: Nit: Any chance you could replace `function_scope` with an `enum`? This would make things a bit more readable for callers.

Also, if it's a  boolean, I'd rather call it `is_function_scope`.

crates/binjs_es6/src/scopes.rs
```diff
@@ -55,16 +52,21 @@ impl AnnotationVisitor {
...
-        for name in free_names_in_current_function.drain() {
+    fn pop_free_names(&mut self, bindings: &[&HashSet<String>], is_leaving_function_scope: bool) {
+        let mut free_names_in_current_block = self.free_names_in_block_stack.pop().unwrap();
+        for (name, old_cross_function) in free_names_in_current_block.drain() {
             let is_bound = bindings.iter()
                 .find(|container| container.contains(&name))
                 .is_some();
             if !is_bound {
-                self.free_names_in_nested_functions.insert(name);
+                if let Some(mut parent_free) = self.free_names_in_block_stack.last_mut() {
```
Yoric: Nit: A comment explaining what you're doing here would be nice.

PR #135, opened 2018-05-24
Normalize binjs_io::multipart::write import section for readability.

kannanvijayan:
I'm starting to read through this code and get an understanding of it.  I found that the import usage in most of the files made it very hard to track down where definitions were coming from.  I'm going to try to make every import name explicit, and avoid '*' imports.

I plan on cleaning this up a bit at a time as I run across it.  This is the first one.

### DIFF
--- a/crates/binjs_io/src/multipart/write.rs
+++ b/crates/binjs_io/src/multipart/write.rs
@@ -1,23 +1,35 @@
+use TokenWriterError;
+
 use bytes;
-use bytes::compress::*;
-use bytes::varnum::*;
-use io::*;
-use ::TokenWriterError;
-use multipart::*;
+use bytes::compress::Compression;
+use bytes::compress::CompressionResult;
+use bytes::varnum::WriteVarNum;
+use io::TokenWriter;
+
+use multipart::FormatInTable;
+use multipart::HEADER_GRAMMAR_TABLE;
+use multipart::HEADER_STRINGS_TABLE;
+use multipart::HEADER_TREE;
 
 use std;
-use std::collections::{ HashMap, HashSet };
+use std::collections::HashMap;
+use std::collections::HashSet;
 use std::cell::RefCell;
-use std::fmt::{ Debug, Display, Formatter };
+use std::fmt::Debug;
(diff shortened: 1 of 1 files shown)


crates/binjs_io/src/multipart/write.rs
```diff
@@ -1,23 +1,35 @@
+use TokenWriterError;
+
 use bytes;
-use bytes::compress::*;
-use bytes::varnum::*;
-use io::*;
-use ::TokenWriterError;
-use multipart::*;
+use bytes::compress::Compression;
+use bytes::compress::CompressionResult;
```
Yoric:
```rust
use bytes::compress::{ Compression, CompressionResult };
```

crates/binjs_io/src/multipart/write.rs
```diff
@@ -1,23 +1,35 @@
...
+use bytes::varnum::WriteVarNum;
+use io::TokenWriter;
+
+use multipart::FormatInTable;
```
Yoric:
```rust
use multipart::{ FormatInTable, HEADER_GRAMMAR_TABLE, HEADER_STRING_TABLE, HEADER_TREE };
```

crates/binjs_io/src/multipart/write.rs
```diff
@@ -1,23 +1,35 @@
...
+use multipart::HEADER_GRAMMAR_TABLE;
+use multipart::HEADER_STRINGS_TABLE;
+use multipart::HEADER_TREE;
 
 use std;
-use std::collections::{ HashMap, HashSet };
```
Yoric: Please don't change the coding style.

kannanvijayan: Np.  Wasn't really thinking of it - just mostly committing the changes I was making to the import list as I tried to read through and understand the code.  I'll make sure to keep the existing multiple-name-import style.

crates/binjs_io/src/multipart/write.rs
kannanvijayan: How do I actually accept these change requrests?

PR #139, opened 2018-06-08
Add binjs_dump command which dumps multipart .binjs file tree structure

arai-a:
Added `binjs_dump` command which dumps the `[TREE]` section structure, along with the hex-dump for corresponding part.
currently this supports `multipart` format only.

here's output for .binjs file which contains `print(10);`:

```
Reading.
Attempting to decode as multipart.

08                      # Script {
                        # .scope
02                      # AssertedVarScope : None
                        # .directives
00                      # list (length=0) []
                        # .statements
02                      # list (length=1) [
04                      # ExpressionStatement {
                        # .expression
00                      # CallExpression {
                        # .callee
06                      # IdentifierExpression {
                        # .name
00                      # string="print"
                        # }
                        # .arguments
02                      # list (length=1) [
0a                      # LiteralNumericExpression {
                        # .value
00 00 00 00 00 00 24 40 # float=10
                        # }
                        # ]
                        # }
                        # }
                        # ]
                        # }
```

any suggestion, especially to the output format, is welcome :)


crates/binjs_es6/src/io.rs
```diff
@@ -67,10 +67,18 @@ impl<R> Deserialization<R, String> for Deserializer<R> where R: TokenReader {
 impl<R, T> Deserialization<R, Vec<T>> for Deserializer<R> where R: TokenReader, Self: Deserialization<R, T> {
     fn deserialize(&mut self) -> Result<Vec<T>, R::Error> {
         let (len, guard) = self.reader.list()?;
+        if len > 0 {
+            dump_format!(self.reader, "list (length={}) [", len);
```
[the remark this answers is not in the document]
Yoric: If it's a debug log, could you use the macros from crate `log`?

crates/binjs_generate_library/src/lib.rs
```diff
@@ -963,10 +970,23 @@ impl<R> Deserialization<R, Option<{name}>> for Deserializer<R> where R: TokenRea
...
+                        .fields()
+                        .iter()
+                        .map(|field| format!("
+        dump_format!(self.reader, \".{name}\");
+        let data_{name} = (self.deserialize() as Result<{spec}, R::Error>)?;
+",
+                            name = field.name().to_rust_identifier_case(),
+                            spec = match field_specs.get(field.name()) {
+                                Some(s) => s,
+                                None => "???"
```
Yoric: Could you document that `"???"`?

crates/binjs_io/src/io.rs
```diff
@@ -93,8 +93,32 @@ pub trait TokenReader where Self::Error: Debug + From<::TokenReaderError>,
     /// read (in particular that all bytes were consumed). In most
     /// implementations, failure to do so will raise an assertion.
     fn untagged_tuple(&mut self) -> Result<Self::UntaggedGuard, Self::Error>;
+
+    // Functions for file format dump.
```
Yoric: What's a dump, exactly?

crates/binjs_io/src/io.rs
```diff
@@ -93,8 +93,32 @@ pub trait TokenReader where Self::Error: Debug + From<::TokenReaderError>,
...
+
+    // Enables/disables dump.
```
Yoric: Nit: `///` if it's documentation.

crates/binjs_io/src/io.rs
Yoric: Also, it looks like it's going to make the implementation of this trait more complicated. I suspect that providing default noop implementations would be useful, what do you think?

crates/binjs_generate_library/src/lib.rs
```diff
@@ -1,3 +1,6 @@
+//#[macro_use]
+extern crate binjs_io;
```
Yoric: Do you need the dependency outside of generated code?

crates/binjs_es6/src/io.rs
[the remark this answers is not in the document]
arai-a: it's a macro to print the given file format comment in the right side of binary representation,
(that handles column position and newlines internally)
which is defined in crates/binjs_io/src/io.rs
I guess I should use more descriptive name.

crates/binjs_es6/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #![feature(box_patterns)]
 #![recursion_limit="128"] // We have deeply nested data structures...
 
+#[macro_use]
```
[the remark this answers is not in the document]
arai-a: I put the above macro inside crates/binjs_io/src/io.rs, since I cannot find any other place to put it, to make it available everywhere which needs it.
let me know if you know better place :)
(I'm rust beginner and I think I'm missing something)

crates/binjs_generate_library/src/lib.rs
arai-a: I'll replace it with unwrap() call, which I learned just yesterday :)

crates/binjs_generate_library/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["David Teller <D.O.Teller@gmail.com>"]
 
 [dependencies]
 binjs_meta = { path = "../binjs_meta/", version = "*" }
+binjs_io = { path = "../binjs_io/", version = "*" }
```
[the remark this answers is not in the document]
arai-a: I'll remove.

crates/binjs_generate_library/src/lib.rs
arai-a: Looks like it was unnecessary.

arai-a:
Addressed the review comments, and also added the description to crates/binjs_io/src/io.rs.
Separated the newly added methods into FileStructurePrinter trait, which has default implementation.

arai-a:
Fixed documentation comments, and also added structural interpretation comment for offset (used by skip)

arai-a:
given that the implementation will likely throw error with offset, now I feel it's better printing the offset as well.

Cargo.toml
```diff
@@ -41,6 +41,10 @@ path = "src/bin/encode.rs"
 name = "binjs_decode"
 path = "src/bin/decode.rs"
 
+[[bin]]
```
Yoric: While you're at it, now that we start to have many binaries, could you add a one-line comment explaining what it does?

crates/binjs_es6/src/lib.rs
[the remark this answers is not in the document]
Yoric: If you're using it both in `binjs_generate_library` and at runtime, it should go in `binjs_shared`. If it's only runtime, `binjs_io` is good.

I should write a README.md explaining this :)

crates/binjs_generate_library/src/lib.rs
```diff
@@ -921,8 +921,11 @@ impl<R> Deserializer<R> where R: TokenReader {{
 
 impl<R> InnerDeserialization<R, {name}> for Deserializer<R> where R: TokenReader {{
     fn deserialize_inner(&mut self) -> Result<{name}, R::Error> where R: TokenReader {{
+        print_file_structure!(self.reader, \"{name} {{{{\");
```
Yoric: Yep, that's nicer.

crates/binjs_io/src/multipart/read.rs
```diff
@@ -142,11 +142,80 @@ impl Deserializer for NodeDescriptionDeserializer {
     }
 }
 
+struct DumpCursor {
```
Yoric: Nit: Could you document what this is for?

src/bin/dump.rs
```diff
@@ -0,0 +1,46 @@
...
+use std::fs::*;
+use std::io::*;
+
+use clap::*;
+
+fn main() {
+    env_logger::init();
+
+    let matches = App::new("BinJS dumper")
+        .author("David Teller, <dteller@mozilla.com>,\n\
```
Yoric: I'm pretty sure you're the main author :)

crates/binjs_generate_library/src/lib.rs
arai-a: The macro is used only in runtime.
I'll keep it inside binjs_io.

Yoric:
(you should be able to merge whenever you're ready)

PR #144, opened 2018-06-29
Update idl to match ecmascript-binary-ast f2bbb2a69fa0d37a36efa70716cf7d6e5fe10851

arai-a:
to be clear, I don't intend do land this immediately.
discussion is here: https://github.com/orgs/binast/teams/binast-core/discussions/7

this updates the in-tree idl and corresponding parts to match to the spec at the point of Jun 28.
https://github.com/binast/ecmascript-binary-ast/commits/spec-draft

The main changes are:
 - AssertedDeclaredName/AssertedBoundName + AssertedDeclaredKind for scopes
 - Skippable interface to Lazy field
   - skip/skipped implicit fields are removed, and {fieldname}_skip field is inserted before the lazy field, which has offset
 - separate Function and Contents
 - separate ArrowExpression for FunctionBody and Expression
 - scope is no more optional
   - when converting from Shift's JSON, dummy scope object is created
 - version number is bumped

### DIFF
--- a/crates/binjs_io/src/multipart/write.rs
+++ b/crates/binjs_io/src/multipart/write.rs
@@ -537,7 +537,7 @@ impl TreeTokenWriter {
             .map_err(TokenWriterError::WriteError)?;
         self.statistics.uncompressed_bytes += MAGIC_HEADER.len();
 
-        const FORMAT_VERSION : u32 = 0;
+        const FORMAT_VERSION : u32 = 1;
         self.data.write_varnum(FORMAT_VERSION)
             .map_err(TokenWriterError::WriteError)?;
         self.statistics.uncompressed_bytes += std::mem::size_of_val(&FORMAT_VERSION);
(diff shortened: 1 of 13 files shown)


arai-a:
the patch for SpiderMonkey is here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1472103

Yoric:
Sorry, I missed this. It won't merge, could you double-check?

arai-a:
@efaust , @Yoric , can you review this?
(I'm going to change the spec maybe shortly, but I think it worth updating binjs-ref to current spec, so that the subsequent update will be more smooth)

arai-a:
to be clear, this doesn't solve https://github.com/binast/binjs-ref/issues/147
and I'm about to fix it at the same time as addressing https://github.com/binast/ecmascript-binary-ast/pull/52

arai-a:
oops, I accidentally pushed to wrong branch. please ignore. I'll revert to the previous commit

crates/binjs_io/src/multipart/write.rs
```diff
@@ -780,61 +780,53 @@ impl TokenWriter for TreeTokenWriter {
...
+            debug!(target: "multipart", "tagged tuple index: {index:?}",
+                index = index);
+
             // Now, we determine if we should add an Offset
-            let inject_in_offset =
-                if let Some(&(_, ref tree)) = children.get(0)
-                {
-                    if let UnresolvedTree {
+            let mut children_data = vec![];
+            let mut is_lazy = false;
```
Yoric: The role of `is_lazy` isn't clear to me. You're trying to detect the field that comes just after a lazy field?

crates/binjs_meta/src/spec.rs
```diff
@@ -106,6 +106,7 @@ pub struct Field {
     name: FieldName,
     type_: Type,
     documentation: Option<String>,
+    is_lazy: bool,
```
Yoric: I believe it would be less error-prone to have an enum `Laziness` instead of a boolean.

I put something along these lines [in that patch](https://github.com/binast/binjs-ref/commit/8a875327360e847276fe793332bf21d6c73b56ec).

crates/binjs_io/src/multipart/write.rs
arai-a: lazy field is split into 2 children in the tree, the offset to skip the field, and the actual field.
then, the loop is trying to detect the offset field, and treat the next field as lazy field.

renamed `is_lazy_field`, and added comments.

PR #145, opened 2018-07-02
Add ToStr.newline_if_not_empty

arai-a:
Corresponds to the function added in https://bugzilla.mozilla.org/show_bug.cgi?id=1472101 , ported to binjs_meta/src/util.rs
I'll remove the one in m-c once this is merged and released.

### DIFF
--- a/crates/binjs_meta/src/util.rs
+++ b/crates/binjs_meta/src/util.rs
@@ -8,6 +8,16 @@ pub trait ToStr {
     fn newline(&self) -> String {
         format!("{}\n", self.to_str())
     }
+
+    /// Append newline if the string is not empty.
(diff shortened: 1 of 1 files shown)


crates/binjs_meta/src/util.rs
```diff
@@ -8,6 +8,14 @@ pub trait ToStr {
     fn newline(&self) -> String {
         format!("{}\n", self.to_str())
     }
+    /// Append newline if the string is not empty.
+    fn newline_if_not_empty(&self) -> String {
```
Yoric: Could you cache the result of `self.to_str()`? It can be rather expensive, in some cases.

PR #149, opened 2018-08-01
Fix the field order in Visitor (fixes #148)

### LINKED ISSUE #148: Visitor visits interface field in the wrong order
the order of how Visitor visits field doesn't match to the spec interface's order.
this is problematic for accumulating binding names.

arai-a:
Changed it to use vector of tuple, instead of HashMap


Yoric:
Merge as you see fit.

PR #158, opened 2018-08-31
Add length field to functions (fixes #153)

### LINKED ISSUE #153: Add function length field for functions
corresponds to https://github.com/binast/ecmascript-binary-ast/pull/55

arai-a:
corresponds to binast/ecmascript-binary-ast#55
dummy_parameter_scope is modified to return both scope and length fields.

### DIFF
--- a/src/source/shift.rs
+++ b/src/source/shift.rs
@@ -195,6 +195,11 @@ enum FunctionKind {
     Setter,
 }
 
+struct ParameterScopeAndFunctionLength {
+    scope: json::JsonValue,
+    length: usize
+}
+
 /// A data structure designed to convert from Shift AST to BinJS AST.
 struct FromShift;
 impl FromShift {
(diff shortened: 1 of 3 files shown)


src/source/shift.rs
```diff
@@ -229,9 +229,10 @@ impl FromShift {
         JSON::Object(scope)
     }
 
-    fn dummy_parameter_scope<'a, I>(&self, params: I) -> json::JsonValue
+    fn parameter_scope_and_length<'a, I>(&self, params: I) -> (json::JsonValue, json::JsonValue)
```
Yoric: I'd be more comfortable if instead of a pair of `JsonValue` you had a `struct` with clearly-labeled names. Or maybe return a `usize` for the length and convert it to `json::JsonValue` at the call site?

arai-a: okay, I'll add struct with usize.

Issue #166, opened 2018-09-28
Add command-line filter mode for encode/decode 


arai-a:
while creating automation for creating testcases, it's better having a filter mode:

the current possible usage is like following:
  cat some.js | babel --presets=env | encode > some.binjs

we also need to have silent option which stops printing progress, or maybe activate the silent mode for filter mode.

PR #167, opened 2018-09-28
Add quiet mode

arai-a:
Just added quiet mode, as a preparation for https://github.com/binast/binjs-ref/issues/166

### DIFF
--- a/src/bin/decode.rs
+++ b/src/bin/decode.rs
@@ -13,6 +13,14 @@ use std::io::*;
 
 use clap::*;
 
+macro_rules! progress {
+    ($quiet:expr, $($args:tt)*) => {
+        if !$quiet {
+            println!($($args)*);
+        }
+    }
+}
+
 fn main() {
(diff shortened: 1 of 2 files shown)


src/bin/decode.rs
```diff
@@ -41,18 +46,24 @@ fn main() {
     // Setup.
     let printer = Shift::new();
 
-    println!("Reading.");
+    if !quiet {
```
Yoric: Maybe make this a macro?

Yoric:
I don't know the best practices here, is it to remove all output or to redirect it to stderr?

arai-a:
given progress is not error, stdout should be fine on non-quiet mode.
on quiet mode, no progress should be printed.

actually, I think even error should not be printed on quiet mode (tell the result only with exit status), but I'm not sure how to handle that in rust's `expect()` etc

Yoric:
Ok, then land as you see fit :)

PR #168, opened 2018-09-28
Add command-line filter mode (fixes #166) 

### LINKED ISSUE #166: Add command-line filter mode for encode/decode

arai-a:
I'm not sure how to handle patch stack on github, anyway here's the patch for command-line filter mode, on https://github.com/binast/binjs-ref/pull/167

### DIFF
--- a/src/bin/decode.rs
+++ b/src/bin/decode.rs
@@ -21,6 +21,18 @@ macro_rules! progress {
     }
 }
 
+// Command line options.
+struct Options<'a> {
+    // True if --print-json is specified.
+    print_json: bool,
+
+    // True if --quiet or -q is specified.
+    quiet: bool,
+
+    // The OUTPUT path, or None if not specified.
+    dest_path: Option<&'a str>,
+}
+
 fn main() {
     env_logger::init();
 
@@ -29,11 +41,9 @@ fn main() {
         .about("Decode a JavaScript BinJS source to a JavaScript text source.")
         .args(&[
             Arg::with_name("INPUT")
-                .required(true)
-                .help("Input file to use. Must be a BinJS source file."),
+                .help("Input file to use. Must be a BinJS source file. If not specified, stdin is used"),
             Arg::with_name("OUTPUT")
-                .required(true)
-                .help("Output file to use. Will be overwritten."),
+                .help("Output file to use. Will be overwritten. If not specified, stdout is used"),
             Arg::with_name("dump")
                 .long("dump")
                 .takes_value(false)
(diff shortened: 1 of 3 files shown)


arai-a:
forgot to explain the usage.

if `--in` and `--out` are not specified, source is read from stdin, and binjs is printed to stdout.
`--quiet` option is implicitly enabled

arai-a:
this implements for "binjs_encode"
I'll prepare another one for decode

arai-a:
fixed decode as well.
if INPUT and OUTPUT are not specified, stdin and stdout are used.

Yoric:
I'll review this once the previous patch has landed, if it's ok.

arai-a:
yes, I'll update this after rebase

arai-a:
now we can do turnaround like this, even with babel :)

```
  cat ${JS_FILE} \
      | ${PATH_TO_BABEL}/babel --presets=${PATH_TO_BABEL_MODULES}/babel-preset-env \
      | (cd ${PATH_TO_BINJS_REF}; ./target/debug/binjs_encode) \
      | (cd ${PATH_TO_BINJS_REF}; ./target/debug/binjs_decode)
```

Yoric:
(waiting for rebase)

src/bin/decode.rs
```diff
@@ -45,40 +49,44 @@ fn main() {
...
+        Some(path) => {
+            parse_tree(&|| {
+                let file = File::open(path)
+                    .expect("Could not open source");
+                BufReader::new(file)
+            }, &options)
+        }
+        None => {
+            let mut buffer = Vec::new();
+            stdin().read_to_end(&mut buffer)
```
Yoric: Looks like you need to go through many hoops because we need to implement `Seek` for our input stream. I seem to remember that `Seek` is only ever used for error messages, so could you try and see if there is a way to get rid of that dependency?

This would simplify your code, and it's not a big deal if we lose a few offsets in error messages at this stage.

If you want to land this before getting rid of `Seek`, that would also work, but in that case, could you file an issue?

src/bin/decode.rs
```diff
@@ -45,40 +49,44 @@ fn main() {
...
+                .expect("Failed to read from stdin");
+
+            parse_tree(&|| {
+                let mut stream = Cursor::new(Vec::new());
```
Yoric: If you head this way, you probably want `let mut stream = Cursor::new(buffer)`, to avoid copying the entire buffer.

src/bin/decode.rs
```diff
@@ -92,12 +100,40 @@ fn main() {
...
+        None => {
+            stdout().write(source.as_bytes())
+                .expect("Could not write destination file");
+        }
+    }
+}
+
+fn parse_tree<R: Read + Seek>(get_stream: &Fn() -> R, options: &Options) -> binjs::specialized::es6::ast::Script
+{
+    if !options.quiet {
```
Yoric: Note for later: we'll eventually need to implement a way to sniff various formats a little bit more nicely.

src/bin/encode.rs
```diff
@@ -102,13 +102,35 @@ fn handle_path<'a>(options: &mut Options<'a>,
...
     progress!(options.quiet, "Parsing.");
-    let json = options.parser.parse_file(source_path)
-        .expect("Could not parse source");
+
+    handle_path_or_text(options, Some(source_path), None,
+                        dest_bin_path, dest_txt_path);
+}
+
+fn handle_path_or_text<'a>(options: &mut Options<'a>,
+    source_path: Option<&Path>,
```
Yoric: Four out of the five arguments look near exchangeable. Could you create a `struct` for all these arguments? Or make them part of `Options`?

src/bin/encode.rs
```diff
@@ -102,13 +102,35 @@ fn handle_path<'a>(options: &mut Options<'a>,
...
+    let (source_len, json) = match source_path {
+        Some(path) => {
+            (std::fs::metadata(path)
+                 .expect("Could not open source")
+                 .len(),
+             options.parser.parse_file(path)
+                 .expect("Could not parse source"))
+        }
+        None => {
+            let text = source_text
```
Yoric: Sounds like you need a

```rust
enum Source {
  FromFile { path: Path },
  FromStdin { text: String }
}
```

or something such.

src/bin/encode.rs
```diff
@@ -188,14 +215,12 @@ fn main_aux() {
...
-                .required(true)
-                .help("Input files to use. Must be JS source file. May be specified multiple times"),
+                .help("Input files to use. Must be JS source file. May be specified multiple times. If not specified, stdin is used."),
             Arg::with_name("out")
                 .long("out")
                 .short("o")
                 .takes_value(true)
-                .required(true)
-                .help("Output directory to use. Files in this directory may be overwritten."),
+                .help("Output directory to use. Files in this directory may be overwritten. Requires --if. If not specified, stdout is used"),
```
Yoric: `--if`?

src/bin/encode.rs
```diff
@@ -257,8 +287,16 @@ fn main_aux() {
         quiet
     };
 
-    for source_path in sources {
-        handle_path(&mut options, source_path, PathBuf::new().as_path());
+    if sources.len() == 0 {
+        let mut buffer = String::new();
+        stdin().read_to_string(&mut buffer)
```
Yoric: So, here, too, why do we block until we have read to the end? Is it just to know the length of the file? If so, I'm not sure we need it all that much.

src/bin/encode.rs
```diff
@@ -257,8 +287,16 @@ fn main_aux() {
         quiet
     };
 
-    for source_path in sources {
-        handle_path(&mut options, source_path, PathBuf::new().as_path());
+    if sources.len() == 0 {
```
Yoric: Could you add a comment to explain that we switch to stdin if there are no sources?

src/bin/decode.rs
arai-a: yes, I'll file a bug, to keep the diff small and scoped.

src/bin/encode.rs
arai-a: yeah, just to keep the same behavior.  I'll file another issue to remove that part from the progress message.

PR #169, opened 2018-09-28
Allow everything in a function expression to shadow the name binding.

efaust:


crates/binjs_es6/src/scopes.rs
```diff
@@ -826,24 +826,24 @@ impl Visitor<()> for AnnotationVisitor {
     }
 
     fn enter_function_expression_contents(&mut self, path: &Path, _node: &mut FunctionExpressionContents) -> Result<VisitMe<()>, ()> {
-        self.push_var_scope(path);
-        self.push_param_scope(path);
-        self.push_this_captured();
         if let Some(_) = self.function_expression_name() {
             self.push_function_name_captured();
         }
+        self.push_var_scope(path);
```
Yoric: Since this is obviously important (and since I wrote that code and even I can't remember why), could you add a comment explaining why this needs to happen after the `if let` above?

crates/binjs_es6/src/scopes.rs
```diff
@@ -826,24 +826,24 @@ impl Visitor<()> for AnnotationVisitor {
...
+        self.push_param_scope(path);
+        self.push_this_captured();
         Ok(VisitMe::HoldThis(()))
     }
     fn exit_function_expression_contents(&mut self, path: &Path, node: &mut FunctionExpressionContents) -> Result<Option<FunctionExpressionContents>, ()> {
+        node.is_this_captured = self.pop_this_captured();
```
Yoric: Same here.

PR #174, opened 2018-10-02
Order table entries with same number of use by data itself (fixes #171)

### LINKED ISSUE #171: sort GRAMMAR table and STRINGS table items
looks like the order of items in GRAMMAR table and STRINGS table is random.
it should be sorted in consistent order (maybe just alphabetically)

arai-a:
Defined the order between the table entries with the same number of uses,
so that the order between them is fixed in the file

### DIFF
--- a/crates/binjs_io/src/multipart/write.rs
+++ b/crates/binjs_io/src/multipart/write.rs
@@ -97,7 +97,7 @@ impl Serializable for Option<SharedString> {
 /// An entry in an WriterTable.
 ///
 /// This entry tracks the number of instances of the entry used in the table.
-struct TableEntry<T> where T: Clone + std::fmt::Debug { // We shouldn't need the `Clone`, sigh.
+struct TableEntry<T> where T: Clone + std::fmt::Debug + Ord { // We shouldn't need the `Clone`, sigh.
     /// Number of instances of this entry around.
     instances: RefCell<u32>,
 
@@ -107,7 +107,7 @@ struct TableEntry<T> where T: Clone + std::fmt::Debug { // We shouldn't need the
     /// The index, actually computed in `write()`.
     index: TableIndex<T>
 }
-impl<T> TableEntry<T> where T: Clone + std::fmt::Debug {
+impl<T> TableEntry<T> where T: Clone + std::fmt::Debug + Ord {
     fn new(data: T) -> Self {
         TableEntry {
             instances: RefCell::new(1),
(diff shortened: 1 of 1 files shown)


crates/binjs_io/src/multipart/write.rs
```diff
@@ -107,22 +107,31 @@ struct TableEntry<T> where T: Clone + std::fmt::Debug { // We shouldn't need the
...
+impl<T> TableEntry<T> where T: Clone + std::fmt::Debug + Ord {
     fn new(data: T) -> Self {
         TableEntry {
             instances: RefCell::new(1),
             index: TableIndex::new(&format!("{:?}", data)),
             data,
         }
     }
+
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
```
Yoric: Looks like you should rather implement `Ord` for `TableEntry<T> where T: Ord`.

crates/binjs_io/src/multipart/write.rs
```diff
@@ -107,22 +107,31 @@ struct TableEntry<T> where T: Clone + std::fmt::Debug { // We shouldn't need the
...
+        match u32::cmp(&*other.instances.borrow(), &*self.instances.borrow()) {
```
Yoric: Turns out that this pattern is so common that `Ordering` has a method `then_with` to factorize it :)

So, something along the lines of:

```rust
u32::cmp(...)
  .then_with(|| self.data.cmp(&other.data)
```

crates/binjs_io/src/multipart/write.rs
```diff
@@ -173,13 +182,13 @@ impl<Entry> WriterTable<Entry> where Entry: Eq + Hash + Clone + Serializable + F
...
-impl<Entry> Serializable for WriterTable<Entry> where Entry: Eq + Hash + Clone + Serializable + FormatInTable + Debug {
+impl<Entry> Serializable for WriterTable<Entry> where Entry: Eq + Hash + Clone + Serializable + FormatInTable + Debug + Ord {
     fn write<W: Write>(&self, out: &mut W) -> Result<usize, std::io::Error> {
         let mut total = 0;
 
-        // Sort entries by number of uses.
+        // Sort entries by number of uses and entry data specific ordering
         let mut contents : Vec<_> = self.map.values().collect();
-        contents.sort_unstable_by(|a, b| u32::cmp(&*b.instances.borrow(), &*a.instances.borrow()));
+        contents.sort_unstable_by(|a, b| a.cmp(b));
```
Yoric: If you implement `Ord` above, this becomes just `contents.sort_unstable()`.

PR #179, opened 2018-10-03
Fix advanced subsubcommand (fixes #178)

### LINKED ISSUE #178: advanced subcommand doesn't handle subsubcommand in binjs_encode
running binjs_encode with advanced expanded uses multipart.
there's bug in from_matches.

arai-a:
given prividers are subsubcommand, we should first check the `advanced` subcommand there and check against `matches` for the subcommand.

### DIFF
--- a/crates/binjs_io/src/lib.rs
+++ b/crates/binjs_io/src/lib.rs
@@ -96,6 +96,8 @@ pub mod xml;
 
 mod util;
 
+const ADVANCED_COMMAND: &str = "advanced";
+
 /// A strategy for placing the dictionary.
 #[derive(Clone, Debug)]
 pub enum DictionaryPlacement {
@@ -393,7 +395,7 @@ impl Format {
 
     /// Return all existing format providers, to manage
     /// command-line arguments.
-    pub fn providers() -> [&'static FormatProvider; 4] {
+    fn providers() -> [&'static FormatProvider; 4] {
         [
             &multipart::FormatProvider,
             &simple::FormatProvider,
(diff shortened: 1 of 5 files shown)


crates/binjs_io/src/lib.rs
```diff
@@ -414,14 +414,16 @@ impl Format {
     /// `matches` as a subcommand. If none, pick the default
     /// provider, without any command-line arguments.
     pub fn from_matches(matches: &clap::ArgMatches) -> Result<Self, std::io::Error> {
-        for provider in Self::providers().into_iter() {
-            let subcommand = provider.subcommand();
-            let key = subcommand.get_name();
-            if let Some(matches) = matches.subcommand_matches(key) {
-                return provider.handle_subcommand(Some(matches));
+        if let Some(matches) = matches.subcommand_matches("advanced") {
```
Yoric: Mmmh... this puts half of `advanced` in encode.rs and half here.

Would it be possible to also make `fn providers()` private and replace it with a `pub fn subcommand() -> SubCommand` that would specify `"advanced"` from this file?

crates/binjs_io/src/lib.rs
```diff
@@ -408,20 +408,30 @@ impl Format {
         &multipart::FormatProvider
     }
 
+    /// Returns command-line arguments.
```
Yoric: Nit: Could you mention that advanced flags are hidden behind a command `"advanced"`?

Yoric: Also, could you turn this `"advanced"` into a named constant?

PR #182, opened 2018-10-04
Entropy with dictionary builder

Yoric:
This patch introduces a module that can build a dictionary from a bunch of files (it's actually connected with parsing only in a test). This replicates a subset of @kannanvijayan's work which is missing to perform efficient entropy coding in the Rust implementation.

Adding @kannanvijayan in case he has something to say about this implementation.


crates/binjs_io/src/entropy/mod.rs
```diff
@@ -55,19 +55,20 @@
 //! and/or custom dictionary later.
 
 pub mod model;
+
 mod predict;
-mod tree;
-pub mod read;
-pub mod write;
+// mod tree;
```
arai-a: I wonder what happens to this file...

crates/binjs_io/src/entropy/model.rs
```diff
@@ -1,183 +1,251 @@
...
 
 use std;
-use std::rc::Rc;
+use std::collections::HashMap;
+
+pub type IOPath = binjs_shared::ast::Path<InterfaceName, (/* child index */ usize, /* field name */ FieldName)>;
+
+#[derive(Debug, Default)]
+pub struct Dictionary<T> {
+    /// All syntax-level booleans, predicted by path.
```
arai-a: can you add comment to somewhere that explains what "syntax-level" is?

crates/binjs_io/src/entropy/model.rs
```diff
@@ -1,183 +1,251 @@
...
+    max_path_depth: usize,
+
+    /// A dictionary.
+    dictionary: &'a mut Dictionary</* instances */ usize>,
+
+    /// Number of instances of each string in the current file.
+    instances_of_strings_in_current_file: StringsInFile<usize>,
+
+    /// Number of files in which each string appears.
+    files_containing_string: &'a mut StringsInFile<usize>,
```
arai-a: can you explain why it's StringsInFile type ?
also, where is the target of reference supposed to be exist ?

crates/binjs_io/src/entropy/model.rs
```diff
@@ -1,183 +1,251 @@
...
+            instances_of_strings_in_current_file: StringsInFile::default(),
+            files_containing_string
+        }
+    }
+
+    fn add_instance_to_path<V>(max_path_depth: usize, value: V, path: &IOPath, source: &mut PathPredict<V, usize>)
+        where
+            V: 'a + std::hash::Hash + Eq + Clone + std::fmt::Debug,
+    {
+        let len = usize::min(max_path_depth, path.len()) + 1;
```
arai-a: with max_path_depth is parameter here, instead of some constant?

crates/binjs_io/src/entropy/model.rs
```diff
@@ -1,183 +1,251 @@
...
+        for len in 0 .. len {
```
arai-a: it would be nice to briefly explain what actually this is doing, in comment.

crates/binjs_io/src/entropy/mod.rs
Yoric: It has become irrelevant. I have upstreamed all the important parts into the auto-generated ast.rs.

crates/binjs_io/src/entropy/model.rs
Yoric: > can you explain why it's StringsInFile type ?

This should be better now.

>  also, where is the target of reference supposed to be exist ?

Er... anywhere upstack?

crates/binjs_io/src/entropy/model.rs
Yoric: Because I'm planning to make it a command-line argument.

crates/binjs_io/src/entropy/model.rs
```diff
@@ -1,183 +1,275 @@
...
+        + self.string_enum_by_path.len()
+        + self.property_key_by_path.len()
+        + self.identifier_name_by_path.len()
+        + self.interface_name_by_path.len()
+        + self.string_literal_by_path.len()
+        + self.list_length_by_path.len()
     }
 }
 
+/// Strings grouped per kind.
```
arai-a: to my understanding, this struct is actually "a map from kind+string to something (T)".
that means, "string" is part of key, not value.

and the actual usage is the following 2 cases:
  * T = the number of appearance of kind+string in single file
  * T = the number of files that has kind+string, among all files

so, "StringsPerKind" still doesn't reflect the actual meaning.
given fields have "instances", how about "InstancesPerKindAndString" ?

crates/binjs_io/src/entropy/model.rs
arai-a: looks like, `add_instance_to_path` and `add_instance_to_strings` are defined like static method, but actually receives `DictionaryBuilder` fields.
can we just make them instance method and receive `self` and extract fields from it ?
that way it's clearer what this method does.

crates/binjs_io/src/entropy/model.rs
Yoric: What about `KindedStringMap`? That sounds a bit more readable.

arai-a: I'm not sure "Kinded" means what we need, but if it means, I think it's good.

Yoric:
> looks like, add_instance_to_path and add_instance_to_strings are defined like static method, but actually receives DictionaryBuilder fields.
> can we just make them instance method and receive self and extract fields from it ?
> that way it's clearer what this method does.

What do you mean "extract fields"?

But generally, the problem is once you have borrowed a bit of `self` (e.g. to make a pointer to `&mut self.foo.bar`), you cannot call methods from `self`, because that would cause unpredictable aliasing.

crates/binjs_io/src/entropy/model.rs
Yoric: Well, I think it can be accepted with that meaning :)

arai-a:
those parameters are expected to be fields of DictionaryBuilder, not something else,
and the method reads fields and modifies fields, it's most likely an instance method.
if there's some reason that it cannot be instance method, can you add comments that what those parameters should be, and why it's static method?

Yoric:
> those parameters are expected to be fields of DictionaryBuilder, not something else,
and the method reads fields and modifies fields, it's most likely an instance method.

If the name of the field was hardcoded, this would work, but that would defeat the entire purpose of this method. I could turn it into a macro, I suppose.

> if there's some reason that it cannot be instance method, can you add comments that what those parameters should be, and why it's static method?

Sure.

Yoric:
Also, note that I'm improving type-safety in a followup commit.

PR #193, opened 2018-10-23
Use WTF-8 as string encoding in binjs file and escape lone surrogate while processing (fixes #190)

### LINKED ISSUE #190: Do not decode escape sequence
String in `var s = "<\uD83D>";` (all ASCII) is encoded as `"<\xEF\xBF\xBD>"`, which means the escape sequence is parsed/decoded and re-encoded as UTF-8, and the code itself becomes `U+FFFD` (given it's invalid)
it shouldn't decode and the encoded string is also `"<\uD83D>"`

arai-a:
Applied the following changes:

  * introduces `escaped_wtf8` module, which converts between WTF-8 and UTF-8 + escape sequence (`\x7F` + `XXXX`)
  * escape the generated JSON in Node side in `shift.rs` (we could output WTF-8 from Node side and decode/escape WTF-8 in Rust side, but I didn't want to add more module dependency in node side)
    * this means all string is escaped while processing in Rust side
      (but of course no effect if the input has no lone surrogate)
  * unescape while writing string to mutlipart file
    * which effectively changed the string encoding of multipart binjs file to WTF-8
    * this needs the change in JS engine side (I have patch, will post shortly)
  * escape while reading string from multipart file
  * added formatter (`escaped_wtf8::for_print`) for `binjs_dump` and used there, in order to see the escaped string in (incorrect but) more similar-to-usual way


arai-a:
patch for SpiderMonkey side https://bugzilla.mozilla.org/show_bug.cgi?id=1501155

Yoric:
@SimonSapin has agreed to review this patch. Let's see how I can add him to the mix.

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -0,0 +1,276 @@
...
+                    LEAD_SORROGATE_MIN,
+                    TRAIL_SORROGATE_MAX);
+
+            const LEAD_SURROGATE_MIN: u16 = 0xD800;
+            let diff = codepoint - LEAD_SURROGATE_MIN;
+            let high = (diff / 64) as u8;
+            let low = (diff % 64) as u8;
+            buf.push(LONE_SORROGATE_UNIT_1);
+            buf.push(LONE_SORROGATE_UNIT_2_MIN + high);
+            buf.push(LONE_SORROGATE_UNIT_3_MIN + low);
```
SimonSapin: Similar to in `escape`, I think this is correct but I half-expected the usual UTF-8 algorithm `0x80 | ((codepoint >> 6) & 0x3F)` and `0x80 | (codepoint & 0x3F)`

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    Some(buf)
+}
+
+/// If the given `bytes` contains any escaped lone surropgate, convert it to
+/// \uXXXX format, just for debug print.
+/// This does unrecoverable conversion, the result shouldn't be used outside
+/// of debug print.
+pub fn for_print(s: &SharedString) -> SharedString {
+    let pos = s.as_str().find(char::from(LONE_SORROGATE_ESCAPE_CHAR));
```
SimonSapin: Since `SharedString` implements `Deref` there’s no need to call `.as_str()` and `s.find(…)` should work. (Method call syntax auto-derefs.)

crates/binjs_io/src/multipart/write.rs
```diff
@@ -55,25 +56,35 @@ impl Serializable for Vec<u8> {
...
+/// - number of WTF-8 bytes (varnum);
+/// - sequence of WTF-8 bytes.
 impl Serializable for SharedString {
     fn write<W: Write>(&self, out: &mut W) -> Result<usize, std::io::Error> {
         let mut total = 0;
-        total += out.write_varnum(self.len() as u32)?;
-        out.write_all(self.deref().as_bytes())?;
+        let bytes = self.deref().as_bytes();
+
+        match escaped_wtf8::unescape(bytes) {
```
SimonSapin: Since you’re writing a generic `W: fmt::Write` stream, maybe `unescape` could be generic and take `&mut W` as well to avoid allocating at all?

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -0,0 +1,276 @@
...
+        }
+
+        // ...... \x7F XXXX ...... \x7F XXXX ......
+        //                  ^      ^
+        //                  |      |
+        //                begin   end
+        end = pos.unwrap();
+    };
+
+    SharedString::from_string(String::from_utf8(buf).expect("Escaped string should be valid UTF-8"))
```
SimonSapin: I think this UTF-8 check could be avoided by making `buf` a `String` instead of `Vec<u8>`, using `push_str` instead of `extend_from_slice`, etc.

crates/binjs_io/src/multipart/write.rs
arai-a: do you mean changing the unescape function to unescape_and_write or something?

crates/binjs_io/src/escaped_wtf8.rs
arai-a: it makes `input[...]` part complicated, and different from `unescape`, so I'd like to keep it as is.

crates/binjs_io/src/multipart/write.rs
SimonSapin: Maybe you’d want to rename it at the same time, but I meant that this `&mut W` parameter would be instead of the return value, not in addition to.

However that’s perhaps a more disruptive change than is worth doing speculatively before profiling to see if those memory allocation take any significant time at all.

crates/binjs_io/src/escaped_wtf8.rs
SimonSapin: Sorry, what `input[...]` part? `str` can be sliced with range indexing syntax too. Anyway, as you prefer. These are just suggestions, as I said before feel free to ignore any :)

PR #197, opened 2018-11-09
Add weird bytes to container header

### LINKED ISSUE #196: Add newline characters to container header
The proposed [BinAST container format](https://github.com/binast/binjs-ref/issues/13) starts with the word "BINAST", then a version number, then the rest is the payload. The PNG format, for comparison, has a much more complicated [header](https://en.wikipedia.org/wiki/Portable_Network_Graphics#File_header):

> Values | Purpose
> -- | --
> 89 | Has the high bit set to detect transmission systems that do not support 8 bit data and to reduce the chance that a text file is mistakenly interpreted as a PNG, or vice versa.
> 50 4E 47 | In ASCII, the letters PNG, allowing a person to identify the format easily if it is viewed in a text editor.
> 0D 0A | A DOS-style line ending (CRLF) to detect DOS-Unix line ending conversion of the data.
> 1A | A byte that stops display of the file under DOS when the command type has been used—the end-of-file character.
> 0A | A Unix-style line ending (LF) to detect Unix-DOS line ending conversion.

I think that BinAST should include similar "check bytes" in its header, particularly the newline characters (to detect if it got line-ending converted) and the `1A` byte (not because we're worried about DOS, but because normal text files never contain that byte, it helps tools like `grep` avoid accidentally interpreting a binast file as ASCII, especially since they will often get served with a `.js` extension).

zealousidealroll:
Changes the stamp at the the beginning of BinAST files to one based on the PNG header.
This is done to make it play nicer when used with software that does text processing:

* Because the stamp contains a NUL byte, most software (particularly Git) will recognize that it is not a text file. The header is also not valid UTF-8, in case any software tries to check text validity that way.
* Because the stamp contains a UNIX newline and a DOS newline, anything that unconditionally performs line conversion will be detected right away.

Fixes #196

### DIFF
--- a/crates/binjs_io/src/multipart/read.rs
+++ b/crates/binjs_io/src/multipart/read.rs
@@ -232,11 +232,15 @@ pub struct TreeTokenReader {
 impl TreeTokenReader {
     pub fn new<R: Read + Seek>(mut reader: R) -> Result<Self, TokenReaderError> {
         // Check magic headers.
-        const MAGIC_HEADER: &'static [u8; 5] = b"BINJS";
+        const MAGIC_HEADER: &'static [u8; 8] = b"\x89BJS\r\n\0\n";
+        const OLD_MAGIC_HEADER: &'static [u8; 5] = b"BINJS";
         const FORMAT_VERSION: u32 = 1;
 
-        reader.read_const(MAGIC_HEADER)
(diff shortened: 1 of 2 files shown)


Yoric:
(I'm going to close this PR)

PR #198, opened 2018-11-09
Disable attempts at lexcial function declarations

efaust:
Since we're not doing lexicals for the initial es5 implementation, don't try to make function declarations into lexicals. We don't properly compute the full AnnexB semantics anyway.

### DIFF
--- a/crates/binjs_es6/src/scopes.rs
+++ b/crates/binjs_es6/src/scopes.rs
@@ -922,7 +922,8 @@ impl Visitor<()> for AnnotationVisitor {
 
         // 1. If the declaration is at the toplevel, the name is declared as a `var`.
         // 2. If the declaration is in a function's toplevel block, the name is declared as a `var`.
-        // 3. Otherwise, the name is declared as a `let`.
+        // 3. Otherwise, the name would be declared as a `let`, subject to proper Annex B requirements.
+        //      NB: Since we don't current support Lexical bindings, we treat all functions as Var..
         debug!(target: "annotating", "exit_eager_function_declaration sees {:?} at {:?}", node.name.name, path.get(0));
         match path.get(0).expect("Impossible AST walk") {
             &WalkPathItem { field: ASTField::Statements, interface: ASTNode::Script } |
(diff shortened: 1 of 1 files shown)


efaust:
This should unblock the outstanding bugs for function bindings.

Yoric:
Looks good to me, but @arai-a is the JS semantics expert.

crates/binjs_es6/src/scopes.rs
```diff
@@ -922,7 +922,8 @@ impl Visitor<()> for AnnotationVisitor {
 
         // 1. If the declaration is at the toplevel, the name is declared as a `var`.
         // 2. If the declaration is in a function's toplevel block, the name is declared as a `var`.
-        // 3. Otherwise, the name is declared as a `let`.
+        // 3. Otherwise, the name would be declared as a `let`, subject to proper Annex B requirements.
+        //      NB: Since we don't current support Lexical bindings, we treat all functions as Var..
```
arai-a: binjs-ref does support lexical bindings, and what we're dropping support for lexical binding is SpiderMonkey's side implementation, at this point.
so, at least the comment doesn't match to the current situation of binjs-ref.

I wonder if we should treat "non-const lexical" as "var" in SpiderMonkey's side.
or maybe completely drop (disable) support for lexical bindings in binjs-ref.

arai-a: (discussed in IRC)
in case the lexical handling in binjs-ref is not fully correct, and this makes the handling compatible with ES5, this solution seems to be fine.
I'd like to suggest slightly changing the wording
  "support Lexical bindings" => "fully support Lexical bindings"
so that it's clearer that there's partial support but the detail is different.

efaust:
As discussed on IRC, holding off on merging this until we have confirmation of Annex B.3.3 semantics from Shu.

Yoric:
@efaust What should we do with this PR?

Yoric:
Closing PR. @efaust, don't hesitate to reopen!

PR #201, opened 2018-11-14
Entropy strings

Yoric:
This time, shorter PR :)

### DIFF
--- a/crates/binjs_io/src/entropy/predict.rs
+++ b/crates/binjs_io/src/entropy/predict.rs
@@ -17,9 +17,21 @@ pub type IOPathItem = binjs_shared::ast::PathItem<InterfaceName, (/* child index
 #[derive(Default, Serialize, Deserialize, From, Into, AddAssign, Clone, Copy)]
 pub struct Instances(usize);
 
+/// A newtype for `usize` used to represent an index in a dictionary of values.
+#[derive(Add, Constructor, Eq, PartialEq, Ord, PartialOrd, Clone, Copy, Into, Debug, Hash, Serialize, Deserialize)]
+struct DictionaryIndex(usize);
+
+/// A newtype for `usize` used to represent a reference to a value already encountered.
+///
+/// By convention, `0` is the latest value, `1` the value before, etc.
+#[derive(Constructor, Eq, PartialEq, Ord, PartialOrd, Clone, Copy, Hash, Into, Debug, Serialize, Deserialize)]
+struct BackReference(usize);
+
+
 mod context_information {
     use super::Instances;
     use entropy::probabilities::{ SymbolIndex, SymbolInfo };
+
     use std::collections::HashMap;
     use std::hash::Hash;
 
@@ -72,8 +84,7 @@ mod context_information {
         }
 
         pub fn value_by_symbol_index(&self, index: SymbolIndex) -> Option<&NodeValue> {
-            let index: usize = index.into();
-            self.value_by_symbol_index.get(index)
+            self.value_by_symbol_index.get(Into::<usize>::into(index))
         }
     }
 
@@ -94,10 +105,7 @@ mod context_information {
             let instances: Vec<_> = self
                 .stats_by_node_value
                 .values()
-                .map(|x| {
-                    let x: usize = x.clone().into();
-                    x as u32
-                })
+                .map(|x| Into::<usize>::into(x.clone()) as u32)
                 .collect();
 
             let distribution = std::rc::Rc::new(std::cell::RefCell::new(range_encoding::CumulativeDistributionFrequency::new(instances).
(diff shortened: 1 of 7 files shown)


crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct WindowPredict<NodeValue, Statistics> where NodeValue: Clone + Eq + Hash {
+    /// The window width
+    width: usize,
+
+    /// All the known values.
+    value_by_dictionary_index: Vec<NodeValue>,
+    dictionary_index_by_value: HashMap<NodeValue, DictionaryIndex>,
+
+    latest_values: Vec<NodeValue>,
```
arai-a: Please add comment what the entries are:
  * the order inside this vector
    * which side is the first
  * where the current index points inside the vector
  * what the current index is
  * where "back reference" points

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Serialize, Deserialize)]
+enum WindowPrediction {
+    BackReference(BackReference),
+    DictionaryIndex(DictionaryIndex),
+}
+
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct WindowPredict<NodeValue, Statistics> where NodeValue: Clone + Eq + Hash {
```
arai-a: please add comment what this is, like:
  * what kind of window (what is this for)
  * what the target sequence is
  * where the "width" points
  * how prediction is performed there
  * what the relation between window and dictionary is

after reading the following code, this looks like cache with a vector that holds entries in MRU order,
and width is the max number of cached entry,
and back reference is the index from the MRU entry.
is it correct?

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+    /// The window width
+    width: usize,
+
+    /// All the known values.
+    value_by_dictionary_index: Vec<NodeValue>,
+    dictionary_index_by_value: HashMap<NodeValue, DictionaryIndex>,
+
+    latest_values: Vec<NodeValue>,
+
+    info: ContextInformation<WindowPrediction, Statistics>,
```
arai-a: this also needs comment

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+        WindowPredict {
+            width,
+            value_by_dictionary_index: Vec::with_capacity(1024), // FIXME: Magic number
+            dictionary_index_by_value: HashMap::with_capacity(1024),
+            latest_values: Vec::with_capacity(width),
+            info: ContextInformation::new(),
+        }
+    }
+
+    /// Update current window by moving an index to the first position.
```
arai-a: what does "first position" mean?

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+    fn update_window_by_index(&mut self, index: BackReference) -> Result<BackReference, ()> {
```
arai-a: the method name doesn't sound correct.
what this does is, remove the entries after index-th value, and move the index-th value to 0-th element.
so, at least "by" is misleading (for me it sounds like "+=")

also, what's the reason why the entries after index-th value is dropped?

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+        let as_usize: usize = index.into();
+        if as_usize >= self.latest_values.len() {
+            return Err(());
+        }
+        let ref mut slice = self.latest_values
+            .as_mut_slice()
+            [0..index.into()];
```
arai-a: s/index.into()/as_usize/ maybe?

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+        slice.rotate_right(1);
+        Ok(index)
+    }
+
+    /// Update current window by putting a value to the first position.
```
arai-a: here too, first position is unclear unless it's explained somewhere.

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+    ///
+    /// If the value is already in the window, it is moved to the first position,
+    /// otherwise, it is added. If the resulting window is too large, it is truncated.
+    fn update_window_by_value(&mut self, value: &NodeValue) -> Option<BackReference> {
```
arai-a: maybe `put_value_into_window` or something?

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+            .as_mut_slice()
+            [0..index.into()];
+        slice.rotate_right(1);
+        Ok(index)
+    }
+
+    /// Update current window by putting a value to the first position.
+    ///
+    /// If the value is already in the window, it is moved to the first position,
+    /// otherwise, it is added. If the resulting window is too large, it is truncated.
```
arai-a: please explain about the return value too.

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+            .position(|v| v == value)
+        {
+            return Some(self.update_window_by_index(BackReference(index))
+                .unwrap())
+        }
+        if self.latest_values.len() < self.width {
+            // If the window isn't full yet, simply add the value at start.
+            self.latest_values.insert(0, value.clone());
+        } else {
+            // Otherwise, rotate.
```
arai-a: "rotate" doesn't correctly explains what happens.
this pushes the value in front of the vector and drops the last value.


btw, if this may happen frequently, can't you use `vec_deque` ?
https://doc.rust-lang.org/std/collections/vec_deque/index.html

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+        let number_of_values = self.value_by_dictionary_index.len();
+        let dictionary_index = *self.dictionary_index_by_value.entry(value.clone())
+            .or_insert(DictionaryIndex(number_of_values));
+
+        if dictionary_index == DictionaryIndex(number_of_values) {
+            // We've just inserted `value`.
+            self.value_by_dictionary_index.push(value.clone());
+        } else {
+            // Value was already known.
+            let index : usize = dictionary_index.into();
```
arai-a: won't this hit unused variable on non-debug build?
(I don't know much about debug/non-debug build difference and warning system tho)

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +346,172 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+                    .clone();
+                if let Err(_) = self.update_window_by_index(index) {
+                    return None;
+                }
+                Some(result)
+            }
+        }
+    }
+
+    pub fn by_value_mut(&mut self, value: &NodeValue) -> Option<&mut SymbolInfo> {
```
arai-a: here too can you put what the returned thing is, into the method name?

src/bin/generate_dictionary.rs
```diff
@@ -167,6 +167,14 @@ fn main_aux() {
                     .map(|_| ())
                     .map_err(|e| format!("Invalid number {}", e)))
                 .help("Maximal path length to store in the dictionary."),
+            Arg::with_name("width")
```
arai-a: window-width or cache-width or something?

crates/binjs_io/src/entropy/predict.rs
Yoric: Indeed.

crates/binjs_io/src/entropy/predict.rs
Yoric: Not quite. Method `as_mut_slice` gives a view on a subset of the vector, so we're not dropping anything. In that subset, we replace the last value, then rotate right.

crates/binjs_io/src/entropy/predict.rs
Yoric: @arai-a I'm almost sure that it won't, but I'll admit that I haven't checked.

crates/binjs_io/src/entropy/predict.rs
Yoric: > btw, if this may happen frequently, can't you use `vec_deque` ?
> https://doc.rust-lang.org/std/collections/vec_deque/index.html

This would improve the performance of `window_insert_value` and reduce that of `window_move_to_front`. I actually have no idea which one is most common, but since both of them are bounded by `width`, which should be fairly small, I don't think it's going to be a major difference.

Let's reconvene when we need to profile :)

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +344,192 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+            dictionary_index_by_value: HashMap::with_capacity(1024),
+            latest_values: Vec::with_capacity(width),
+            info: ContextInformation::new(),
+        }
+    }
+
+    /// Update current window by moving the value at index `index` to the latest-seen
+    /// position (index 0).
+    ///
+    /// Fails if the index is not in [0, self.latest_values.len()[
```
arai-a: this has failure branch instead of assertion because the `index` is supposed to be raw value read from external file and can be outside the range, right?

crates/binjs_io/src/entropy/predict.rs
```diff
@@ -336,3 +344,192 @@ impl<NodeValue> PathPredict<NodeValue, SymbolInfo> where NodeValue: Eq + Hash +
...
+        let ref mut slice = self.latest_values
+            .as_mut_slice()
+            [0..as_usize];
+        slice.rotate_right(1);
+        Ok(())
+    }
+
+    /// Update current window by putting a value to the latest-seen position (index 0).
+    ///
+    /// If the value is already in the window, it is moved to the last-seen position,
```
arai-a: "last-seen" => "latest-seen" maybe?

PR #207, opened 2018-11-15
Gathering a few statistics on entropy write

Yoric:


crates/binjs_io/src/entropy/mod.rs
```diff
@@ -76,6 +79,16 @@ pub struct Options {
...
     /// at this stage.
     string_tables: KindedStringMap<SymbolInfo>,
+
+    /// Statistics obtained while writing.
+    /// If several files are written with the same options, we accumulate
+    /// statistics.
+    content_lengths: Rc<RefCell<write::ContentInfo<usize>>>,
+}
+impl Options {
+    pub fn write_statistics(&self) -> Ref<write::ContentInfo<usize>> {
```
arai-a: this sounds like this method writes
can it be renamed to `statistics_for_write` or something?

PR #213, opened 2018-11-16
Trivial unit test for entropy roundtrip

Yoric:


tests/test_dictionary_builder.rs
```diff
@@ -124,6 +136,49 @@ test!(test_model_dictionary_builder, {
         ],
         |value| SharedString::from_string(value.to_string())
     );
+
+    let options = entropy::Options::new(
+        dictionary.instances_to_probabilities("dictionary"),
+        files_containing_string.instances_to_probabilities("strings"),
+    );
+
+    eprintln!("Starting roundtrip with dictionary");
```
arai-a: I guess either this or remaining `println!`s are typo
those lines look supposed to be printed to same descriptor (either stdout or stderr)

Yoric: Fixed, thanks.

PR #215, opened 2018-11-16
Fix: Generated AST library pops interfaces even in case of error

Yoric:


arai-a:
what does "pops" mean here?

Yoric:
The opposite of `push()`: remove latest inserted element.

arai-a:
I don't understand the issue

Yoric:
It's a combination of two things:

- `Path` asserts in a few places that we removed fields and interfaces in the reverse order in which they were added;
- when the serializer encounters an error (e.g. an i/o error), it bails out immediately, without removing fields/interfaces properly;
- by combining the above, we encounter a case in which an error in the dictionary triggers an assertion failure;
- I have tweaked our ast generator to fix this.

Yoric:
Thank you :)

PR #219, opened 2018-11-19
Resolves #218: Get rid of entropy.strings

### LINKED ISSUE #218: In entropy, get rid of dict.strings
These statistics are only meant to determine which strings to keep in the dictionary and which strings to prune. We don't actually need to write this to a file.

Yoric:
This file is unused, let's get rid of it.


Yoric:
Note that we still collect the underlying data, we just don't write it to disk.

arai-a:
forgot to mention.  r+ with the 2nd argument of `entropy::Options::new` removed in the consumer

PR #220, opened 2018-11-20
Moving ContentInfo into its own module

Yoric:
Also, making it a bit more generic, since we're starting to use it a lot.


crates/binjs_io/src/io/content.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+impl<T> std::iter::FromIterator<(T, &'static str)> for ContentInfo<Option<T>> {
+    fn from_iter<U>(iter: U) -> Self
+    where
+        U: IntoIterator<Item = (T, &'static str)>
+    {
+        let mut container = Self::with(|_| None);
+        for (value, name) in iter.into_iter() {
```
arai-a: can you tell me what this block actually does?
this mostly looks nop+assertion but not sure what area/scope `replace` affects

crates/binjs_io/src/io/content.rs
```diff
@@ -0,0 +1,120 @@
...
+			(self.property_keys, "property_keys"),
+			(self.identifier_names, "identifier_names"),
+			(self.interface_names, "interface_names"),
+			(self.string_literals, "string_literals"),
+			(self.list_lengths, "list_lengths"),
+        ].into_iter()
+    }
+}
+
+impl<T> std::iter::FromIterator<(T, &'static str)> for ContentInfo<Option<T>> {
```
arai-a: where is this used, and what is this for?

crates/binjs_io/src/io/content.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+impl<T> ContentInfo<T> {
+    /// Initialize a new `ContentInfo`.
+    pub fn with<F>(f: F) -> Self
+        where F: Fn(&str) -> T
+    {
+        ContentInfo {
+            bools: f("bools"),
+            floats: f("floats"),
+			unsigned_longs: f("unsigned_longs"),
```
arai-a: please replace tab with spaces, here and several places below

crates/binjs_io/src/io/content.rs
Yoric: We replace the content of the field (`container.bools`, `container.floats`, etc.) with the `value` extracted from the iterator.

`replace(...)` returns the old value, which lets us assert that we haven't inserted two values for the same field

crates/binjs_io/src/io/content.rs
Yoric: It deserializes a `ContentInfo` from a data structure. So, typical use

```rust
let mut info = HashMap::new();
// Somehow, fill `info` e.g. by reading a dictionary or a compressed file.

let info2: ContentInfo<_> = info.into_iter()
    .collect();
// Optionally, call `into_with` to convert from a `ContentInfo<Option<T>>` into a `ContentInfo<T>`.

```

crates/binjs_io/src/io/content.rs
arai-a: I still don't understand the semantics (or maybe higher overview of what this method i for)
what does this method actually do? why do you replace?

PR #230, opened 2018-11-23
move to thread rng instead of OsRng for WASM

sendilkumarn:
`lalrpop` has some transitive dependency that still uses `rand 0.5.5` which causes problem while doing WASM compilation.

Also it is good for WASM to use `thread_rng` instead of `OsRng` here.


Yoric:
Could you explain in the commit message why we need this change?

PR #231, opened 2018-11-23
Generic node type

Yoric:


crates/binjs_shared/src/ast.rs
```diff
@@ -202,4 +202,10 @@ impl<I, F> std::borrow::Borrow<[PathItem<I, F>]> for Path<I, F> where I: Debug +
     fn borrow(&self) -> &[PathItem<I, F>] {
         &self.items
     }
-}
\ No newline at end of file
+}
+
+/// The root type for nodes in the AST.
+pub trait Node: downcast_rs::Downcast {
+    fn name(&self) -> &'static str;
```
arai-a: Can you add comment what this method is for, and what the expected return value is?
also, perhaps, how and where downcast is performed, so that it's clearer what this trait is for.

Yoric: I just realized that there may be a slightly simpler manner of doing this, so I'll wait until this is useful to continue.

arai-a: to be clear, I'm fine landing this, with/without this part
(especially, removing obsolete file is nice to do)

PR #237, opened 2018-11-29
Use {enter,exit}_* in and remove Guard from TokenReader.

arai-a:
(possibly conflict with some changes in other PR)

having separate struct to track the end of list/tuple is hard to manage, because of ownership,
and I think just introducing enter/leave is sufficient for existing case.

Removed Guard structs/traits, and replaced it with {enter,leave}_{list,tagged_tuple,untagged_tuple}_at.

let me know entropy is going to introduce another guard and it conflicts with this.

will ask review after test.

### DIFF
--- a/crates/binjs_io/src/io/mod.rs
+++ b/crates/binjs_io/src/io/mod.rs
@@ -162,26 +162,8 @@ pub type Path = binjs_shared::ast::Path<InterfaceName, /* Field */ (usize, Field
 /// which lets the reader determine what item we're reading in the AST. This
 /// may be used both for debugging purposes and for encodings that depend
 /// on the current position in the AST (e.g. entropy coding).
-pub trait TokenReader: FileStructurePrinter where Self::ListGuard: Guard,
-                                                  Self::TaggedGuard: Guard,
-                                                  Self::UntaggedGuard: Guard,
-                                                  Self: Sized
+pub trait TokenReader: FileStructurePrinter where Self: Sized
 {
-    /// A guard, used to make sure that the consumer has properly read a list.
-    ///
-    /// See the documentation of `self.list`.
-    type ListGuard;
-
-    /// A guard, used to make sure that the consumer has properly read a tagged tuple.
-    ///
-    /// See the documentation of `self.tagged_tuple`.
-    type TaggedGuard;
-
-    /// A guard, used to make sure that the consumer has properly read an untagged tuple.
-    ///
-    /// See the documentation of `self.untagged_tuple`.
-    type UntaggedGuard;
-
     /// Poison the reader, ensuring that it will never be used for reading again.
     fn poison(&mut self) {
         // Do nothing.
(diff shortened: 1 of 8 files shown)


Yoric:
You're right, I was planning to remove them, but didn't get the opportunity to do so.

So 👍

crates/binjs_io/src/io/mod.rs
```diff
@@ -239,33 +221,24 @@ pub trait TokenReader: FileStructurePrinter where Self::ListGuard: Guard,
...
-    /// Returns a guard for that list and the number of elements
-    /// in the list. Once the list is read entirely, callers MUST
-    /// call `guard.done()` to ensure that the list was properly
-    /// read (in particular that all bytes were consumed). In most
-    /// implementations, failure to do so will raise an assertion.
-    fn list_at(&mut self, _path: &Path) -> Result<(u32, Self::ListGuard), TokenReaderError>;
+    fn enter_list_at(&mut self, _path: &Path) -> Result<u32, TokenReaderError>;
+
+    /// Finish reading a list.
+    fn leave_list_at(&mut self, _path: &Path) -> Result<(), TokenReaderError>;
```
Yoric: On the writer side, it's called `enter_`/`exit_`, could you keep that naming convention?

Yoric: We probably want an empty default implementation.

crates/binjs_io/src/io/mod.rs
```diff
@@ -239,33 +221,24 @@ pub trait TokenReader: FileStructurePrinter where Self::ListGuard: Guard,
...
-    /// the contents must be read, and a guard for that tuple.
-    /// Once the tuple is read entirely, callers MUST
-    /// call `guard.done()` to ensure that the tuple was properly
-    /// read (in particular that all bytes were consumed). In most
-    /// implementations, failure to do so will raise an assertion.
-    fn tagged_tuple_at(&mut self, _path: &Path) -> Result<(InterfaceName, Option<Rc<Box<[FieldName]>>>, Self::TaggedGuard), TokenReaderError>;
+    fn enter_tagged_tuple_at(&mut self, _path: &Path) -> Result<(InterfaceName, Option<Rc<Box<[FieldName]>>>), TokenReaderError>;
+
+    /// Finish reading a tagged tuple.
+    fn leave_tagged_tuple_at(&mut self, _path: &Path) -> Result<(), TokenReaderError>;
```
Yoric: Same here: name and default implementation.

crates/binjs_io/src/io/mod.rs
```diff
@@ -239,33 +221,24 @@ pub trait TokenReader: FileStructurePrinter where Self::ListGuard: Guard,
...
-    ///
-    /// Once the tuple is read entirely, callers MUST
-    /// call `guard.done()` to ensure that the tuple was properly
-    /// read (in particular that all bytes were consumed). In most
-    /// implementations, failure to do so will raise an assertion.
-    fn untagged_tuple_at(&mut self, _path: &Path) -> Result<Self::UntaggedGuard, TokenReaderError>;
+    fn enter_untagged_tuple_at(&mut self, _path: &Path) -> Result<(), TokenReaderError>;
+
+    /// Finish reading an untagged tuple.
+    fn leave_untagged_tuple_at(&mut self, _path: &Path) -> Result<(), TokenReaderError>;
```
Yoric: Same here: name and default implementation.

Issue #229, opened 2018-11-22
Parse with Ratel


Yoric:
The [Ratel project](https://github.com/ratel-rust/ratel-core) has a really fast JS parser for Rust. It would be nice if we could use it. It's not the same AST, so we'd need conversion code. Although it probably cannot be our default parser, as long as it's not stable and we don't know how well battle-tested it is, but it would clearly be useful for our internal tests.

PR #238, opened 2018-11-30
Add JSON reader/writer and binjs_convert_from_json command (fixes #194)

### LINKED ISSUE #194: add a command to convert possibly-invalid JSON to binjs file
We need a way to create an invalid-file testcase reliably and automatically, which is usable across format version bump, to test implementation against invalid input efficiently.

I think XML is sufficient to represent most possible format-independent invalid structure,
such like SyntaxError or something.

If such XML can be converted to other format without performing error check, that can be used as a testcase.

arai-a:
This add the following:
  - `advanced json` option, which can read/write JSON format of BinAST file
  - `binjs_convert_from_json` command which reads the above JSON file and encode in the other format BinAST file.

`binjs::io::json` contains Decoder, Encoder, JS value, and JSON parser.

I created yet another JSON parser in order to handle escape sequence and lone surrogate,
the code is mostly manual translation of JSONParser in m-c.

The expected usage is the following,
where `some_filter` is a command (maybe written in JS or Python) that mutates the JSON file in order to create invalid content,
like, modifying the tagged tuple field type, or the number of field items, etc.
The final result is the another format of BinAST file with invalid content, which is intended to be used to test decoder (SpiderMonkey impl etc).

```
$ cat test.js \
    | cargo +nightly --run binjs_encode  -- advanced json \
    | some_filter \
    | cargo +nightly --run binjs_convert_from_json > test.binjs
```

This way, such invalid file testcase can be generated only with input `test.js` and `some_filter`,
both are text files, and this is robust to the file format change.

### DIFF
--- a/crates/binjs_io/src/escaped_wtf8.rs
+++ b/crates/binjs_io/src/escaped_wtf8.rs
@@ -1,4 +1,4 @@
-/// An utility to convert between WTF-8 and UTF-8 + special; escape sequence.
+/// An utility to convert between WTF-8 and UTF-8 + special escape sequence.
 ///
 /// The escape sequence has the following syntax:
 ///   \x7F (single delete character) + XXXX (4 hex digits in ASCII)
@@ -29,7 +29,17 @@ const LONE_SURROGATE_UNIT_3_MIN: u8 = 0x80;
 const LONE_SURROGATE_UNIT_3_MAX: u8 = 0xBF;
 
 const LEAD_SURROGATE_MIN: u16 = 0xD800;
+const LEAD_SURROGATE_MAX: u16 = 0xDBFF;
+const TRAIL_SURROGATE_MIN: u16 = 0xDC00;
 const TRAIL_SURROGATE_MAX: u16 = 0xDFFF;
+
+fn is_lead_surrogate(n: u16) -> bool {
+    return LEAD_SURROGATE_MIN <= n && n <= LEAD_SURROGATE_MAX;
+}
+
+fn is_trail_surrogate(n: u16) -> bool {
+    return TRAIL_SURROGATE_MIN <= n && n <= TRAIL_SURROGATE_MAX;
+}
 
 /// Convert 0-F number to ASCII char.
 fn encode_hex_char(n: u8) -> u8 {
@@ -37,6 +47,20 @@ fn encode_hex_char(n: u8) -> u8 {
         0...9 => b'0' + n,
         0xa...0xf => b'A' + (n - 10),
         _ => panic!("unexpected input"),
+    }
+}
+
+/// Returns true if the given code is 0-9A-Fa-f chars.
+/// This uses u16 instead of u8 because the callsite uses u16.
+fn is_hex_char(n: u16) -> bool {
+    if b'0' as u16 <= n && n <= b'9' as u16 {
+        true
+    } else if b'A' as u16 <= n && n <= b'F' as u16 {
+        true
+    } else if b'a' as u16 <= n && n <= b'f' as u16 {
+        true
+    } else {
+        false
     }
 }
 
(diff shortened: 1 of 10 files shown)


Yoric:
Is there any chance that your json parsing code could be somehow submitted as a patch to https://github.com/maciejhirsz/json-rust ?

arai-a:
given we use WTF-8, and we use custom escape, it's hard...

crates/binjs_io/src/json/parser.rs
```diff
@@ -0,0 +1,754 @@
...
+    /// At the start of the parse, before any values have been processed.
+    JSONValue,
+
+    /// After reading a token for any value.
+    JSONValueSwitch(Token),
+}
+
+/// No-op function just for readability.
+/// Designates goto in the original code.
+fn goto(label: NextLabel) -> NextLabel {
```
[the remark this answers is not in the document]
arai-a: in original code it's real `goto`.

```
   A:
     ...
   B:
     ...
     goto A;
```

it's replaced with the following loop,
and goto function is there just to clarify it corresponds to goto.

```
loop {
   next_label = match next_label {
     NextLabel::A => {
       ...
     }
     NextLabel::B => {
       ...
       goto(NextLabel::A)
     }
   };
}
```

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -295,3 +319,222 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+                    "escaped codepoint should be either {:04x} or lone surrogate ({:04x}...{:04x})",
+                    LONE_SURROGATE_ESCAPE_CHAR,
+                    LEAD_SURROGATE_MIN,
+                    TRAIL_SURROGATE_MAX);
+
+            buf.extend_from_slice(b"\\u");
+            buf.push(input[i + 1]);
+            buf.push(input[i + 2]);
+            buf.push(input[i + 3]);
+            buf.push(input[i + 4]);
```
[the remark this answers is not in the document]
arai-a: The input is supposed to be escaped_wtf8 string.
I'll make the comment clearer.

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -295,3 +319,222 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+            } else if next == b'n' as u16 {
+                buf.push(b'\n' as u16);
+                i += 1;
+            } else if next == b'r' as u16 {
+                buf.push(b'\r' as u16);
+                i += 1;
+            } else if next == b't' as u16 {
+                buf.push(b'\t' as u16);
+                i += 1;
+            } else if next == b'u' as u16 {
```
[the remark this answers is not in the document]
arai-a: what do you mean?

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -295,3 +319,222 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+///
+/// This is not optimized and only for testing.
+pub fn from_json(s: &SharedString) -> Option<SharedString> {
+    // To simplify the surrogate pair handling, do the following:
+    //   1. convert to UTF-16,
+    //   2. unescape JSON string escape sequence
+    //   3. convert lone surrogate to special escape sequence
+    //   3. convert to UTF-8
+
+    let input: Vec<u16> = s.as_str().encode_utf16().collect();
```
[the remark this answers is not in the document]
arai-a: What's the condition whether to use multiple lines?

crates/binjs_io/src/json/parser.rs
```diff
@@ -0,0 +1,754 @@
...
+            source,
+            current: 0,
+            is_eof: false,
+            stack: Vec::new(),
+        }
+    }
+
+    /// Read a byte from the stream.
+    /// Returns `true` if it successfully read, and `false` is it reaches EOF.
+    fn get_char(&mut self) -> Result<bool, TokenReaderError> {
```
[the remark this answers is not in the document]
arai-a: Just that I don't want to map errors in many callsites.

crates/binjs_io/src/json/parser.rs
```diff
@@ -0,0 +1,754 @@
...
+
+        let mut buf = [0; 1];
+
+        let size = self.source.read(&mut buf)
+            .map_err(TokenReaderError::ReadError)?;
+        if size == 0 {
+            self.is_eof = true;
+            return Ok(false);
+        }
+        self.current = buf[0];
```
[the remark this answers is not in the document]
arai-a: Maybe the method name isn't good.
The method corresponds to `current++` in the original code,
so just to advance cursor to the next code unit.
I'll think about better name.

also, I'd like to restrict the way to access the current code unit only to`self.current`,
given the same code unit is used across multiple places, and mixing the field and return value seems to be confusing.

crates/binjs_io/src/escaped_wtf8.rs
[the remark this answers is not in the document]
Yoric: After method calls, we go to the next line.

RReverser:
It's the standard implementation of deserialize_bytes: https://docs.serde.rs/serde/de/trait.Deserializer.html#tymethod.deserialize_bytes

RReverser:
If this approach sounds good, I'll be happy to work on and make a PR for serde-based JSON myself. As mentioned above, I have an initial implementation locally, and I believe it should significantly simplify JSON handling in the codebase while not sacrificing (and sometimes even improving) performance.

arai-a:
> One more thing: https://github.com/tc39/proposal-well-formed-stringify fixes JSON.stringify behaviour for lone surrogates in strings, turning them into well-defined escapes rather than outputting invalid chars, so now any JSON itself is going to be valid UTF-8.
> 
> This is already implemented in V8 and SpiderMonkey and should be shipped in Node soon-ish.
> 
> With that in mind, it seems useful to polyfill same standards-compliant behaviour and reuse same escaping mechanism rather than use special sequence `\u007F` for custom treatment.

Escape with `\u007F` isn't used in strinfigied JSON side.
It's used as internal representation for lone surrogate in JSON (which is \uXXXX escaped).

While parsing JSON, we unesdcape all escape sequence (`\r`, `\n`, `\xXX`, `\uXXXX`, etc) inside string.
Then, given we use `str` or `String` in our entire rust code, so simply unescaping lone sorrogate cannot be done because it becomes invalid UTF8 sequence.

The motivation for using `\u007F + XXXX` escape internally is that , I wanted to avoid rewrite everything to use `wtf8` struct, at least while experimenting with new format/compression.
So, using `Vec<u8>` or `[u8]` instead doesn't solve the issue.

The special escape sequence is used only internally, and converted from/to normal escape sequence `\uXXXX` or WTF-8 when reading/writing file.
currently, JSON file is using `\uXXXX` escape, and .binjs file is using WTF-8.

RReverser:
Hmm I've seen that aside from internal representation, it's also used between Node and Rust which makes it look like it's used for preserving lone surrogates when passing JSON to Rust, and that's the supposed standard violation that made me suggest another way. Are you saying that regular JSON with escaped lone surrogates (and not `0x7F`) would work as well?

RReverser:
Another thing that put me off was description of the PR:

> in order to handle escape sequence and lone surrogate

So it does sound like this is the main reason for new parser, and that's where I suggest to use Serde instead which gives much more flexibility over parsing input and would allow to just provide custom unescaping for strings (based on `deserialize_bytes`) rather than having to use a completely custom parser.

RReverser:
> So, using Vec<u8> or [u8] instead doesn't solve the issue.

FWIW I didn't suggest this; I suggested `deserialize_bytes` as just a way to implement custom unescape logic on top of regular string parsing.

arai-a:
yes, Node/Rust interface uses the `\x7F` escape in JSON, but it was not in the scope of this issue (especially when I say "JSON".  which is only about the external file).
maybe we could reuse either mine or yours for both case later tho.

then, about Node/Rust interface, I've used `\x7F` escape because rust-json cannot handle escaped lone surrogate properly, and `\x7F` was quick solution which matches to the requirement in rust side (as mentioned above).
I don't think it's standard violation (it's just that the string is different than the original code), but of course removing such thing, or entire dependency to Node (#229), should be better.

> > in order to handle escape sequence and lone surrogate
> 
> So it does sound like this is the main reason for new parser

Yes.

If you're willing to replace this code with serde-json way, I'm fine.
Feel free to take this over.

RReverser:
I think given the CLI and Encoder/Decoder parts it might be worth merging this as-is and then migrate to serde-json on top of it to make progress easier. What do you think?

arai-a:
> I think given the CLI and Encoder/Decoder parts it might be worth merging this as-is and then migrate to serde-json on top of it to make progress easier. What do you think?

sounds reasonable :)

Yoric:
> If this approach sounds good, I'll be happy to work on and make a PR for serde-based JSON myself. As mentioned above, I have an initial implementation locally, and I believe it should significantly simplify JSON handling in the codebase while not sacrificing (and sometimes even improving) performance.

If this works, that would be very nice.

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -29,8 +29,18 @@ const LONE_SURROGATE_UNIT_3_MIN: u8 = 0x80;
 const LONE_SURROGATE_UNIT_3_MAX: u8 = 0xBF;
 
 const LEAD_SURROGATE_MIN: u16 = 0xD800;
```
Yoric: Could you document these constants and functions?

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -279,3 +303,222 @@ pub fn for_print(s: &SharedString) -> SharedString {
 
     SharedString::from_string(String::from_utf8(buf).expect("Escaped string should be valid UTF-8"))
 }
+
+/// Convert the given escaped_wtf8 string to JSON.  If the given string contains
```
Yoric: Shouldn't this be called `escape`?

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -279,3 +303,222 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+
+/// Parse UTF-16 encoded JSON string and return UTF-16 string,
+/// which can contain raw lone surrogate.
+fn parse_json(input: &[u16]) -> Result<Vec<u16>, ()> {
+    assert!(input.len() >= 2);
+    assert!(input[0] == b'"' as u16);
+    assert!(input[input.len() - 1] == b'"' as u16);
+
+    let mut buf: Vec<u16> = Vec::with_capacity(input.len());
+    let mut i = 1;
```
Yoric: I think I would be more comfortable if you used a `Cursor<&[u16]>` instead of an index `i`.

crates/binjs_io/src/escaped_wtf8.rs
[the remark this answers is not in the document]
Yoric: I meant `\uXXXX` (if I recall the syntax correctly).

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -279,3 +303,222 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+            i += 1;
+        }
+    }
+    Ok(buf)
+}
+
+/// Escape lone surrogate in the given UTF-16 string.
+fn escape_utf16(input: &[u16]) -> Vec<u16> {
+    let mut buf: Vec<u16> = Vec::with_capacity(input.len());
+    let mut i = 0;
```
Yoric: Here, too, I'd be more comfortable with a `Cursor<&[u8]>` instead of an index.

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -279,3 +303,222 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+
+        i += 5;
+    }
+    buf.push(b'\"');
+    SharedString::from_string(String::from_utf8(buf).expect("Escaped string should be valid UTF-8"))
+}
+
+/// Parse UTF-16 encoded JSON string and return UTF-16 string,
+/// which can contain raw lone surrogate.
+fn parse_json(input: &[u16]) -> Result<Vec<u16>, ()> {
```
Yoric: What's the `Result` for?

crates/binjs_io/src/json/value.rs
```diff
@@ -0,0 +1,216 @@
...
+use TokenReaderError;
+
+use binjs_shared::SharedString;
+
+use std::borrow::Borrow;
+use std::collections::HashMap;
+use std::rc::Rc;
+
+/// JS value in JSON.
+pub enum Value {
```
Yoric: It's a bit annoying to have two different implementations of JSON in the code. Could you document in the file the difference between this implementation and the one in json.rs?

crates/binjs_io/src/json/value.rs
```diff
@@ -0,0 +1,216 @@
...
+    String(SharedString),
+    Number(f64),
+    Object(Rc<HashMap<SharedString, Rc<Value>>>),
```
Yoric: You should probably define
```rust
pub type Object = Rc<HashMap<SharedString, Rc<Value>>>;
```

This would make some things easier to read.

crates/binjs_io/src/json/value.rs
```diff
@@ -0,0 +1,216 @@
...
+    Array(Rc<Vec<Rc<Value>>>),
```
Yoric: Likewise with
```rust
pub type Array = Rc<Vec<Rc<Value>>>;
```

crates/binjs_io/src/json/value.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// Fails if the property doesn't exist.
+    fn get_property(&self, key: &'static str) -> Result<Rc<Value>, TokenReaderError>;
+
+    /// Get the property from the object and return bool or null content.
+    /// Returns Some(bool) if the property is bool.
+    /// Returns None if the property is null
+    ///
+    /// Fails if the property doesn't exist, or the property value is neither
+    /// bool nor null.
+    fn get_bool_or_null_property(
```
Yoric: Maybe `get_maybe_bool_property`, for consistency with other code (here and other methods below)?

crates/binjs_io/src/json/value.rs
Yoric: Also, `Value` is read-only. I imagine that's by design, right?

crates/binjs_io/src/json/write.rs
```diff
@@ -0,0 +1,352 @@
+//! A JSON writer for BinAST file structure.
+//! The written file is supposed to be filtered with some command to generate
```
Yoric: So what kind of filter are you expecting, exactly?

crates/binjs_io/src/json/value.rs
Yoric: @arai-a It's my understanding that we don't intend to use this module for complicated code, only for generating test files, right?
If so, we don't care all that much about performance.

crates/binjs_io/src/escaped_wtf8.rs
Yoric: Ah, nevermind. It's self-contained, so it shouldn't really matter.

crates/binjs_io/src/json/value.rs
RReverser: Isn't this going to be used for regular parsing of input from Shift? But also, if we don't care about performance, then I see even less reasons to use `Rc` here.

Yoric: @arai-a, @RReverser and myself have different understanding of the objective of this code. Could you settle the issue? :)

crates/binjs_io/src/escaped_wtf8.rs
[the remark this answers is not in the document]
arai-a: yes, this handles that syntax.
but I don't understand what your initial comment means.
are you pointing some issue?

crates/binjs_io/src/escaped_wtf8.rs
arai-a: this returns string literal representation of the given string, including quotation,
that is compatible with JSON,
so I think `to_json` is more correct than `escape`.

crates/binjs_io/src/json/value.rs
arai-a: yeah, I'll add comment.

crates/binjs_io/src/json/value.rs
arai-a: `Rc` there is just to make it easier to have reference to a certain object in the tree while reading (traversing the tree), and I cannot find another way...
If there's another simpler way without `Rc`, I'm happy to use it.

then, at this point this is only for a testing-purpose-only command.
The implementation will be replaced with serde-json? (@RReverser would know better about the plan I think?)

crates/binjs_io/src/json/write.rs
arai-a: The filter modifies a valid tree into a invalid tree, which can be encoded as a file, but it's expected to result in an error.
Something like, add unexpected field to some interface, remove field, put a subtree which is not allowed by syntax into somewhere.

The file is used to fuzz the decoder implementation, more efficiently than just mutating the binary file,
and also to keep testing specific scenario across file-format version bump.

I'll add comments.

arai-a:
The objective here is nothing related to WTF-8.
and also Shift has nothing to do with this.

arai-a:
This is a solution for https://github.com/binast/binjs-ref/issues/194 .

This is a tool that generates .binjs file that contains possibly-invalid content,
which cannot be achieved by encoding plain .js to .binjs.

the JSON file represents the AST, but more specifically the order of tokens that reader/writer of each file format (multipart, entropy, etc) handles.
so that this has wider range of expressiveness (regarding the binjs file content) than plain .js. 

The .binjs file converted from the JSON file is fed to decoder implementation (mainly JS engine, like SpiderMonkey's token reader/parser) to test the robustness of the implementation
against malicious/crafted .binjs.

this uses JSON format in order to make it possible to keep the testcase working across file-format/spec version bump.
if we use plain binary format like current multipart, the testcase becomes invalid (that means, it doesn't test the specific part which the original one intended) whenever we changed the file format or interface or something.
but as long as we use text format and the filter-way (that is explained in the issue above),
we can generate the testcase that tests the specific scenario with new file format.

then, the reason why I implemented JSON reader/writer here is just that I couldn't find a library that handles WTF-8, which can be the content of .binjs file.
(@RReverser pointed out that serde-json can be used, but that should be addressed in the another issue)

RReverser:
> then, the reason why I implemented JSON reader/writer here is just that I couldn't find a library that handles WTF-8

I guess what @Yoric asks is why current FromJSON is not suitable for that - it already contains mechanism for encoding/decoding lone surrogates, so it should be suitable for generating such test cases, no?

arai-a:
iiuc, FromJSON is a conversion between JSON and valid AST (it uses valid interface nodes),
and its expressiveness is narrower than this.

RReverser:
I see. But it should still be possible to use same `x7F`-based encoding for lone surrogates for tests with invalid AST? It wouldn't be standard, but given that it will be used just for tests, it seems simpler than introduction of a separate parser.

arai-a:
WTF-8 or lone surrogate is *not* the objective here.
it's just a part of requirement.

then, what I need is a tool that can handle invalid AST (invalid sequence of tokens that .binjs file itself can represent, but doesn't map to BinAST spec's AST), or something like that,
to test implementation.

so that I need something that skips BinAST spec's AST handling,
which is, the code that is auto-generated from IDL.

FromJSON doesn't work for that purpose, right?
or am I missing something?

arai-a:
`binjs_generic` still has a certain fixed grammar (even if it's not specific to ES6). this means it cannot be used for random invalid ASTs that .binjs file format itself can represent.

by "can represent", I mean just by calling `TokenWriter` methods in certain order.
this includes interface that has extra field, or lacks some of fields, two or more fields are swapped,
or have list node or number node where tagged tuple node is expected.

arai-a:
(discussed in IRC)
I'll remove the JSON parser and use either serde-json with lone surrogate handling or json-rs with yet-another-escape-sequence for lone surrogate, after some test how they work.

in either way, I'll post PR without lone surrogate handling as a first step.

arai-a:
anyway, I'll open another PR for simplicity.

PR #245, opened 2018-12-17
Fix building on Windows

RReverser:
Please see commit messages for details.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,35 +1,24 @@
-use std::io::ErrorKind;
-use std::process::*;
+extern crate which;
 
-const PATH_PACKAGES : &'static str = "package.json";
+use std::process::Command;
+use which::which;
+
+const PATH_PACKAGES: &'static str = "package.json";
 
 fn main() {
     // Install Node packages if necessary.
     println!("cargo:rerun-if-changed={}", PATH_PACKAGES);
 
-    // First try with `yarn`, if available.
-    println!("Bootstrap: Attempting to install dependencies with yarn.");
-    match Command::new("yarn")
+    let tool =
+        which("yarn")
+        .or_else(|_| which("npm"))
+        .expect("Could not find neither `yarn` nor `npm`");
+
+    let status =
+        Command::new(tool)
         .arg("install")
-        .spawn()
-    {
-        Ok(mut process) => {
-            process.wait()
-                .expect("Bootstrap: Could not complete `yarn install`");
-        }
-        Err(err) => {
-            if let ErrorKind::NotFound = err.kind() {
-                println!("Bootstrap: Yarn not available, attempting to install dependencies with npm.");
-                // Otherwise, try with `npm`.
-                Command::new("npm")
-                    .arg("install")
-                    .spawn()
-                    .expect("Bootstrap: Could not launch either `yarn install` or `npm install`")
-                    .wait()
-                    .expect("Bootstrap: Could not complete `npm` install");
-            } else {
-                panic!("Bootstrap: Unexpected error during yarn/npm install: {:?}", err);
-            }
-        }
-    }
+        .status()
+        .unwrap();
+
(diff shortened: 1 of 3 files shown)


RReverser:
I think I found another place which doesn't work on Windows; feel free to review but hold off merging for now please, I'd like to fix 'em all in the same PR.

RReverser:
Nvm, that other issue was something different.

RReverser:
Updated to use `which` crate instead of guessing the shell and executing command through it, so build script still works on any platform.

RReverser:
@Yoric Ping. This somewhat complicates my other contributions as I have to develop on top of these fixes but then separate out PRs without them :)

Yoric:
Sorry about the delay, I was on vacation. I'll try and review this today.

build.rs
```diff
@@ -1,35 +1,24 @@
...
 
 fn main() {
     // Install Node packages if necessary.
     println!("cargo:rerun-if-changed={}", PATH_PACKAGES);
 
-    // First try with `yarn`, if available.
-    println!("Bootstrap: Attempting to install dependencies with yarn.");
-    match Command::new("yarn")
+    let tool =
+        which("yarn")
```
Yoric: Oh, that's good. I wasn't aware of `which`.

src/source/shift.rs
```diff
@@ -184,7 +184,7 @@ impl SourceParser for Shift {
             var parseScript = require('shift-parser').parseScript;
             var fs      = require('fs');
 
-            var source  = fs.readFileSync('{}', {{encoding: "utf-8"}});
+            var source  = fs.readFileSync({:?}, {{encoding: "utf-8"}});
```
Yoric: Why a debug print?

RReverser: To fix paths. Please see commit message for details :)

Yoric: Oops, lost track of the commit message :)

Could we use `String::escape_default` or `String::escape_unicode` here? I believe that this would be clearer.

RReverser: I guess, this just felt simpler to add because it also includes quotes and is shorter.

build.rs
```diff
@@ -1,35 +1,26 @@
...
-                    .wait()
-                    .expect("Bootstrap: Could not complete `npm` install");
-            } else {
-                panic!("Bootstrap: Unexpected error during yarn/npm install: {:?}", err);
-            }
-        }
-    }
-}
\ No newline at end of file
+    exec("yarn install").or_else(|()| exec("npm install"))
```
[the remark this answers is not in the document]
RReverser: > Nit: In recent code, we adopted the style

This code is autoformatted with cargo fmt rather than manually, and I think you agreed that it would be good to use it rather than enforce manual style?

build.rs
RReverser: Me neither before this :)

src/source/shift.rs
RReverser: Actually this will be immediately removed in the "reuse Node" PR so I guess not much point in adjusting specifics?

build.rs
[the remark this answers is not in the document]
RReverser: As for debug messages, I'm not sure they're very useful here because stderr is now already inherited, so both errors and install messages will be already printed.

src/source/shift.rs
Yoric: Fair enough :)

build.rs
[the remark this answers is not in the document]
Yoric: > This code is autoformatted with cargo fmt rather than manually, and I think you agreed that it would be good to use it rather than enforce manual style?

Ah, right,good point.



> As for debug messages, I'm not sure they're very useful here because stderr is now already inherited, so both errors and install messages will be already printed.

Ah, right.

PR #247, opened 2019-01-03
Performance: reuse Node.js instance for Shift tasks

RReverser:
This series of commits first simplifies parsing benchmarking to a trivial case to more reliably measure just the overhead of Rust -> Node.js interaction since it's the main contributor to performance issues when invoking Shift tasks.

Then, it changes previous model of running Node.js with given script as an input for each task, to spawning long-term Node.js "services" instead, with each script living in a separate file.

Before Node.js would just accept and spit either raw string or JSON. Now, all the communication is done via ND-JSON (newline-delimited JSON) streams, using the stdin/stdout pipes of the spawned process.

Also, before parsing or other errors would kill the process, dumping the error stack into stderr. This is fine for a single-run model, but not acceptable for long-living processes, so now instead each result of is wrapped into Rust's `Result`-like JSON enum `{"type":"Ok","value":...}` or `{"type":"Err","value":...}` and decoded back from that.

On the above mentioned benchmark, this brings 6x-7x improvement to a parsing speed on my machine, and the improvement is quite noticeable when running `cargo test` as well.

### DIFF
--- a/src/source/parse_str.js
+++ b/src/source/parse_str.js
@@ -1,0 +1,8 @@
+var parseScript = require('shift-parser').parseScript;
+var getStdin = require('get-stdin');
+var escape = require('./escape_wtf8.js');
+
+getStdin().then(data => {
+	var parsed = parseScript(data, { earlyErrors: false });
+	process.stdout.write(escape(JSON.stringify(parsed)));
+});
--- a/src/source/parse_str.js
+++ b/src/source/parse_str.js
@@ -1,8 +1,6 @@
-var parseScript = require('shift-parser').parseScript;
-var getStdin = require('get-stdin');
-var escape = require('./escape_wtf8.js');
+'use strict';
 
-getStdin().then(data => {
-	var parsed = parseScript(data, { earlyErrors: false });
-	process.stdout.write(escape(JSON.stringify(parsed)));
-});
+const { parseScript } = require('shift-parser');
+const startJSONStream = require('./start-json-stream');
+
+startJSONStream(code => parseScript(code, { earlyErrors: false }));
--- a/src/source/parse_str.js
+++ b/src/source/parse_str.js
@@ -1,3 +1,8 @@
+/**
+ * This CLI "daemon" expects JSON strings containing JavaScript code on the
+ * stdin and will parse them and produce Shift AST objects as JSON on stdout.
+ */
+
 'use strict';
 
 const { parseScript } = require('shift-parser');
--- a/src/source/shift.rs
+++ b/src/source/shift.rs
@@ -85,6 +85,14 @@ impl Script {
         })))
     }
 
+    /// This function is responsible for piping JSON inputs to the script
+    /// in a newline-delimited format and parsing JSON outputs back.
+    ///
+    /// Each output can be either `{"type":"Ok","value":...}` if operation was
+    /// successful or `{"type":"Err","value":...}` if corresponding JS callback
+    /// has failed with an error, and these are converted into Rust `Result`.
+    ///
+    /// See start-json-stream.js for some more details.
     pub fn transform(&self, input: &JSON) -> Result<JSON, Error> {
         let output = (move || {
             let mut io = self.0.lock().unwrap();
--- a/src/source/shift.rs
+++ b/src/source/shift.rs
@@ -44,35 +44,7 @@ impl Shift {
         }
     }
 
-    fn parse_script_output(&self, script: &str) -> Result<String, Error> {
-        debug!(target: "Shift", "Preparing script {}", script);
-
-        let script = format!(
-            r##"
-        var result;
-        try {{
-            result = (function() {{
-                {}
-            }})();
-        }} catch (ex) {{
-            console.warn(ex);
-            /* rethrow */ throw ex;
-        }};
-        var process = require('process');
-        /* See crates/binjs_io/src/escaped_wtf8.rs */
-        result = result
-            .replace(/[\u007F\uD800-\uDFFF]/ug, function(m) {{
-                if (m == "\u007F") {{
-                    return "\u007F007F";
-                }}
-                return "\u007F" + m.charCodeAt(0).toString(16);
-            }});
-        process.stdout.write(result);
-        console.warn(result);
-        "##,
-            script
-        );
-
+    fn parse_script_output(&self, script: &str, data: &str) -> Result<String, Error> {
         debug!(target: "Shift", "Launching script {}", script);
 
         let node_memory = match env::var("NODE_MAX_OLD_SPACE_SIZE") {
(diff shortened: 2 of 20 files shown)


src/source/shift.rs
```diff
@@ -4,141 +4,136 @@ use json;
 use json::object::Object;
```
Yoric: While the changes that you're making are a great improvement, I imagine that in some cases, the additional escaping cost will be non-trivial. Perhaps we should document this somewhere?

src/source/shift.rs
```diff
@@ -4,141 +4,136 @@ use json;
...
-        script);
-
-        debug!(target: "Shift", "Launching script {}", script);
-
-        let node_memory = match env::var("NODE_MAX_OLD_SPACE_SIZE") {
-            Err(_) => String::from("--max_old_space_size=2048"),
-            Ok(v) => format!("--max_old_space_size={}", v)
-        };
+struct ScriptIO {
+    input: LineWriter<ChildStdin>,
```
Yoric: As usual, doc would be nice :)

src/source/parse_str.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+const { parseScript } = require('shift-parser');
+const startJSONStream = require('./start-json-stream');
+
+startJSONStream(code => parseScript(code, { earlyErrors: false }));
```
Yoric: How does all of this interact with @arai-a's patch on WTF-8?

src/source/shift.rs
```diff
@@ -4,141 +4,136 @@ use json;
...
+            input: LineWriter::new(child.stdin.unwrap()),
+            output: BufReader::new(child.stdout.unwrap()).lines(),
+        })))
+    }
 
-        debug!(target: "Shift", "Output {:?}", output);
-        if !output.status.success() {
-            error!(target: "Shift", "Script error {}", String::from_utf8(output.stderr).unwrap());
-            return Err(Error::ReturnedError(output.status));
+    pub fn transform(&self, input: &JSON) -> Result<JSON, Error> {
```
Yoric: Transform what into what?

Again, documentation would be useful :)

src/source/shift.rs
```diff
@@ -4,141 +4,136 @@ use json;
...
+        let output = (move || {
+            let mut io = self.0.lock().unwrap();
+            input.write(&mut io.input)?;
+            writeln!(io.input)?;
+            io.output.next().unwrap()
+        })().map_err(Error::IOError)?;
+
+        let result = json::parse(&output).map_err(Error::JsonError)?;
+        if let JSON::Object(mut obj) = result {
```
Yoric: What is this? Error-handling?

src/source/shift.rs
```diff
@@ -4,141 +4,136 @@ use json;
...
-        Ok(result)
+        Err(Error::JsonError(json::Error::wrong_type("Result-like JSON object")))
     }
+}
 
-    fn parse_script_json_output(&self, script: &str) -> Result<JSON, Error> {
-        let stdout = self.parse_script_output(script)?;
+/// Using a Node + Shift binary to parse an AST.
+pub struct Shift {
+    parse_str: Script,
```
Yoric: If I read this correctly, we have three processes launched whenever we call `try_new`, right?

Could you document this? Also, as an optimization, we could probably create them lazily.

src/source/shift.rs
```diff
@@ -4,141 +4,136 @@ use json;
...
+            parse_str: Script::try_new(&node, "parse_str.js")?,
+            parse_file: Script::try_new(&node, "parse_file.js")?,
+            codegen: Script::try_new(&node, "codegen.js")?,
+        })
     }
 
-    pub fn to_source(&self, syntax: &Spec, ast: &JSON) -> Result<String, Error> {
-        let mut ast = ast.clone();
-
+    pub fn to_source(&self, syntax: &Spec, mut ast: JSON) -> Result<String, Error> {
```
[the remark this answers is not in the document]
RReverser: FWIW `mut` is not really part of API - it's needed internally so that Rust would know we can mutate that JSON, but for consumer and in docs it's still the same take-by-value ownership semantics.

src/source/shift.rs
RReverser: I guess we could, but for now this is in the worst case taking same time as before the PR (when one invocation == one reencoded script), and in the best case you create parser just once and process lots of scripts anyway, and then process creation time is now negligible.

src/source/shift.rs
RReverser: Yeah. It's ugly, but will get better with Serde too.

src/source/shift.rs
RReverser: Anything into anything (as all scritpts are different, this is just a generic API to pass some input to them and get some output back).

src/source/parse_str.js
RReverser: This still uses his original `encodeWTF8` function as before, so it shouldn't change anything.

src/source/shift.rs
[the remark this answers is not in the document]
Yoric: Sure. It's the ownership transfer that needs to be documented.

src/source/parse_str.js
Yoric: Good!

src/source/shift.rs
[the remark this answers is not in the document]
RReverser: Ah ok.

src/source/shift.rs
Yoric: Sure. Just noticing a low-hanging fruit :)

src/source/shift.rs
RReverser: What do you mean by the additional escaping cost? The only case I can think of where something new is happening is when passing strings to the codegen.

src/source/shift.rs
Yoric: (don't hesitate to make this a followup issue)

src/source/shift.rs
Yoric: There's one more level of string escaping as there used to be, no? If not, ignore my comment :)

RReverser: No, in most places escaping is the same as before. It always used to call `encodeWTF8` on top of `JSON.stringify` just like now, except it was all hidden inside a JS wrapper whereas now this is done in separate modules.

src/source/shift.rs
RReverser: I don't really know what to add to this one TBH.

I doubt stating "This is lines of input" and "This is lines of output" would be useful given that it's what field names and types already say, so it would be helpful to know what you'd expect here :)

RReverser:
@Yoric Addressed most of the comments.

PR #252, opened 2019-01-14
Migrate from webidl to weedle

RReverser:
This should significantly improve compilation times because `webidl` has `lalrpop` as a dependency which is pretty heavy and blows them up, whereas `weedle` uses just `nom` and so parses with pretty much pure Rust.

Some reference results from my machine with `cargo clean && cargo build -j4 [--release]`:

|         | `webidl` | `weedle` | diff    | diff (%) |
| ------- | -------- | -------- | ------- | -------- |
| debug   | 2m 34s   | 2m 01s   | -0m 33s | -21%     |
| release | 5m 43s   | 4m 37s   | -1m 06s | -19%     |

Initially inspired by https://github.com/rustwasm/wasm-bindgen/pull/639 which performed a similar migration.

### DIFF
--- a/crates/binjs_meta/src/import.rs
+++ b/crates/binjs_meta/src/import.rs
@@ -1,22 +1,33 @@
-use spec::{self, Laziness, SpecBuilder, TypeSum};
+use spec::{self, Laziness, SpecBuilder, TypeSpec, TypeSum};
+use weedle::common::Identifier;
+use weedle::types::*;
+use weedle::*;
 
-use webidl::ast::*;
+fn nullable<T: std::fmt::Debug>(src: &MayBeNull<T>, dst: TypeSpec) -> spec::Type {
+    if src.q_mark.is_some() {
+        dst.optional()
+            .unwrap_or_else(|| panic!("This type could not be made optional {:?}", src.type_))
+    } else {
+        dst.required()
+    }
+}
 
 pub struct Importer {
     builder: SpecBuilder,
     /// The interfaces we have traversed so far.
     path: Vec<String>,
 }
+
 impl Importer {
     /// Import an AST into a SpecBuilder.
     ///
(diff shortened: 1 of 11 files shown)


Yoric:
Does this impact package size? We've had problems with `lalrpop` as a dependency, but it might be a dependency of several components.

RReverser:
`lalrpop` is no longer in the dependency tree (otherwise we wouldn't see a compilation speed improvement either).

Which package size are you interested in? I can take a look.

RReverser:
Actually, rlib sizes and exe sizes didn't change (as this is a build dependency) so I'm not sure what else to check.

crates/binjs_meta/src/import.rs
```diff
@@ -62,19 +73,21 @@ impl Importer {
...
         let mut importer = Importer {
             path: Vec::with_capacity(256),
             builder: SpecBuilder::new(),
         };
         importer.import_ast(ast);
         importer.builder
     }
-    fn import_ast(&mut self, ast: &AST) {
+
+    fn import_ast(&mut self, ast: &Definitions) {
```
Yoric: Maybe `import_all_definitions`?

Yoric:
I was thinking of the size of `cargo package`.

crates/binjs_meta/src/import.rs
RReverser: Renamed.

RReverser:
> I was thinking of the size of cargo package.

Ah, I never used it before. Let me check.

RReverser:
Huh, maybe I'm misusing `cargo package` somehow, but it doesn't print anything for me although seems to be doing something in the background...

RReverser:
Oh, apparently it just took it a long time.

PR #253, opened 2019-01-15
Landing Entropy 0.2 (first part)

Yoric:
Ok, let's start landing entropy 0.2!

### DIFF
--- a/crates/binjs_io/src/bytes/float.rs
+++ b/crates/binjs_io/src/bytes/float.rs
@@ -1,5 +1,7 @@
 use bytes::varnum::*;
 
+use binjs_shared::F64;
+
 use std;
 use std::io::Write;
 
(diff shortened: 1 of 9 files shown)


crates/binjs_io/src/lib.rs
```diff
@@ -18,10 +18,13 @@ extern crate range_encoding;
 #[macro_use]
 extern crate serde_derive;
 extern crate serde;
+extern crate smallvec;
```
arai-a: where is this used?

crates/binjs_io/src/entropy/write.rs
```diff
@@ -9,8 +9,10 @@
 
 use ::TokenWriterError;
 use ::io::{ Path, TokenWriter };
-use ::io::statistics::{ ContentInfo, Instances };
+use ::io::statistics::{ Bytes, BytesAndInstances, ContentInfo, Instances };
 use bytes::lengthwriter::LengthWriter;
+use bytes::varnum::WriteVarNum;
+use super::probabilities::IntoStatistics;
```
arai-a: what are above 3 lines change for?

crates/binjs_io/src/bytes/float.rs
```diff
@@ -90,6 +102,56 @@ impl<T> WriteVarFloat for T where T: Write {
...
+                    return Err(std::io::Error::new(std::io::ErrorKind::InvalidData, "Invalid varfloat"));
+                }
+            }
+        }
+        // This float value maps to a i32.
+        let as_i32 =
+            if as_u32 % 2 == 0 {
+                (as_u32 / 2) as i32
+            } else {
+                - (((as_u32 - 1) / 2) as i32)
```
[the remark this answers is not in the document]
Yoric: I wanted to make sure that numbers between -100 and +100 (which I'm pretty sure are the most common) get encoded in one single byte.

crates/binjs_io/src/lib.rs
Yoric: Might be in a followup patch.

crates/binjs_io/src/bytes/float.rs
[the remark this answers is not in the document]
arai-a: sounds reasonable, but then please add comment to somewhere about the encoding details and the reasons.

PR #262, opened 2019-01-29
Entropy 0.3: Implementing baseline dictionaries

Yoric:
This pull request implements baseline dictionaries, which are used as the extension mechanism.

Followup pull requests will actually make use of these dictionaries.


crates/binjs_io/src/entropy/baseline.rs
```diff
@@ -0,0 +1,229 @@
...
+        path: &IOPath,
+        type_spec: &spec::TypeSpec,
+        or_null: bool,
+        one_level_paths_by_interface_name: &mut HashMap<InterfaceName, HashSet<IOPath>>,
+    ) {
+        use self::spec::NamedType;
+        use self::spec::TypeSpec;
+        match type_spec {
+            TypeSpec::Array { ref contents, .. } => {
+                // Entering an array doesn't affect the path.
```
arai-a: can you explain a bit more what this comment implicitly means?
this sounds like some operation is omitted here, but not clear what it is.

crates/binjs_io/src/entropy/baseline.rs
```diff
@@ -0,0 +1,229 @@
...
+                            self.dictionary
+                                .interface_name_by_path
+                                .add_if_absent(path.borrow(), self.null_name.clone());
+                            one_level_paths_by_interface_name
+                                .entry(self.null_name.clone())
+                                .or_insert_with(|| HashSet::new())
+                                .insert(path.clone());
+                        }
+                    }
+                    NamedType::Typedef(ref def) => self.seed_one_field(
```
arai-a: can you use block here, to make this more similar to other block, for readability?

crates/binjs_io/src/entropy/baseline.rs
```diff
@@ -0,0 +1,229 @@
...
+                }
+            }
+            TypeSpec::NamedType(ref name) => {
+                let named_type = self.spec.get_type_by_name(name).unwrap(); // Grammar has been verified already.
+                match named_type {
+                    NamedType::Interface(ref interface) => {
+                        let interface_name =
+                            InterfaceName::from_rc_string(interface.name().to_rc_string().clone());
+
+                        // Store path => interface and interface => path.
```
arai-a: `"path => interface" and "interface => path".` for readability.

PR #265, opened 2019-01-30
Resolves #263 - Implement Library::spec()

### LINKED ISSUE #263: Generate the entire spec from the webidl and a spec file
Currently, the webidl doesn't contain:

- the name of the root (`"Program"`);
- the name of the `null` interface (`""`).

Consequently, each callsite of the spec needs to specify these. That's error-prone.

We should put these in a spec file, parse it and generate the spec in such a manner that we can instantiate it as a one-liner.

Yoric:
This will simplify some code in Entropy 0.3


crates/binjs_generate_library/src/lib.rs
```diff
@@ -133,15 +133,34 @@ use json::JsonValue as JSON;
...
+        builder.into_spec(SpecOptions {
+            null: &names.meta_null_interface,
+            root: &names.meta_entry_point,
+        })
+    }
+}
+",
+        );
+        impl_buffer.push_str(
+            "impl Library {
```
arai-a: given `spec` and `new` are defined unconditionally, you can use single `impl Library` block for them.

PR #266, opened 2019-01-30
Support escaped WTF-8 in JSON format.

arai-a:
With this patch, the JSON format uses \uXXXX instead of internal escaped WTF-8.

The conversion is done on the entire stringified JSON, not for each string in the tree.

When reading, it first reads string from stream, and then convert \uXXXX with lone surrogate there and replaces them with internal escaped WTF-8 with `from_unicode_escape`, and then parse it.

When writing, it generates the tree and stringifies with raw internal escaped WTF-8, and then convert escaped WTF-8 with \uXXXX with `to_unicode_escape`, and write it.

those 2 operations are lossless as long as the input is in valid encoding, and it's JSON.

then, `to_unicode_escape` does the same thing as pre-existing `for_print`.
so just renamed it and changed the signature (`SharedString` => `String`).
it's still usable for debugging purpose (where `for_print` was used), but in that case the conversion is lossy, because the input is not JSON.

### DIFF
--- a/crates/binjs_io/src/escaped_wtf8.rs
+++ b/crates/binjs_io/src/escaped_wtf8.rs
@@ -1,4 +1,4 @@
-/// An utility to convert between WTF-8 and UTF-8 + special; escape sequence.
+/// An utility to convert between WTF-8 and UTF-8 + special escape sequence.
 ///
 /// The escape sequence has the following syntax:
 ///   \x7F (single delete character) + XXXX (4 hex digits in ASCII)
@@ -18,8 +18,6 @@
 /// in JSON, and most likely unused in actual JS code.
 use std::borrow::Cow;
 
-use binjs_shared::SharedString;
-
 const LONE_SURROGATE_ESCAPE_CHAR: u8 = 0x7F;
 
 const LONE_SURROGATE_UNIT_1: u8 = 0xED;
@@ -29,7 +27,17 @@ const LONE_SURROGATE_UNIT_3_MIN: u8 = 0x80;
 const LONE_SURROGATE_UNIT_3_MAX: u8 = 0xBF;
 
 const LEAD_SURROGATE_MIN: u16 = 0xD800;
+const LEAD_SURROGATE_MAX: u16 = 0xDBFF;
+const TRAIL_SURROGATE_MIN: u16 = 0xDC00;
 const TRAIL_SURROGATE_MAX: u16 = 0xDFFF;
+
+fn is_lead_surrogate(n: u16) -> bool {
+    return LEAD_SURROGATE_MIN <= n && n <= LEAD_SURROGATE_MAX;
+}
+
+fn is_trail_surrogate(n: u16) -> bool {
+    return TRAIL_SURROGATE_MIN <= n && n <= TRAIL_SURROGATE_MAX;
+}
 
 /// Convert 0-F number to ASCII char.
 fn encode_hex_char(n: u8) -> u8 {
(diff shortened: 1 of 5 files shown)


crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -277,5 +283,176 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+    loop {
+        let head = &input[..end];
+        buf.extend_from_slice(head);
+
+        let c = input[end + 1];
+        if c == b'u' {
+            let c1 = input[end + 2];
+            let c2 = input[end + 3];
+            let c3 = input[end + 4];
+            let c4 = input[end + 5];
```
Yoric: Same for `end + 2`...`end + 5`: could you document why panic is reasonable behavior?

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -277,5 +283,176 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+                // If we find lead surrogate, it can be surrogate pair.
+                // Check if there's trailing surrogate t.
+                //
+                // ...... \uXXXX \uXXXX ......
+                //        ^      ^
+                //        |      |
+                //        end    pos
+                let pos = end + 6;
+
+                if input[pos] == b'\\' && input[pos + 1] == b'u' {
```
Yoric: Here, too, could you document why the code will panic?

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -277,5 +283,176 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
     }
 
-    SharedString::from_string(String::from_utf8(buf).expect("Escaped string should be valid UTF-8"))
+    String::from_utf8(buf).expect("Unescaped string should be valid UTF-8")
+}
+
+/// If the given `bytes` contains any unicode escape \uXXXX and that is lone
+/// surrogate, convert it to escaped lone surropgate.
+///
+/// This assumes the input is well-formed JSON, and panics otherwise.
```
Yoric: If I recall correctly, the convention is

```
/// # Failures
///
/// If the input is not well-formed JSON, this function causes a panic.
```

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -277,5 +283,176 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+                //        end    pos
+                let pos = end + 6;
+
+                if input[pos] == b'\\' && input[pos + 1] == b'u' {
+                    let t1 = input[pos + 2];
+                    let t2 = input[pos + 3];
+                    let t3 = input[pos + 4];
+                    let t4 = input[pos + 5];
+
+                    let maybe_tral: u16 = (decode_hex_char(t1) << 12)
```
Yoric: `tral`? Is it a typo on `trail`?

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -277,5 +283,176 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+    let mut input = s.as_str().as_bytes();
+
+    let mut buf: Vec<u8> = Vec::with_capacity(input.len().next_power_of_two());
+    loop {
+        let head = &input[..end];
+        buf.extend_from_slice(head);
+
+        let c = input[end + 1];
+        if c == b'u' {
+            let c1 = input[end + 2];
```
Yoric: Looks like this block is repeated often.

Maybe you should factor out
```rust
let c1 = input[end + 2];
let c2 = input[end + 3];
let c3 = input[end + 4];
let c4 = input[end + 5];
let codepoint: u16 = (decode_hex_char(c1) << 12)
                | (decode_hex_char(c2) << 8)
                | (decode_hex_char(c3) << 4)
                | decode_hex_char(c4);
```

as a function?

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -277,5 +283,176 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+
+                    let maybe_tral: u16 = (decode_hex_char(t1) << 12)
+                        | (decode_hex_char(t2) << 8)
+                        | (decode_hex_char(t3) << 4)
+                        | decode_hex_char(t4);
+                    if is_trail_surrogate(maybe_tral) {
+                        // This is escaped surrogate pair, do not use
+                        // escaped-wtf8.
+
+                        buf.push(b'\\');
```
Yoric: Or
```rust
buf.extend_from_slice(&[b'\\', b'u', c1, c2, c3, c4]);
```

Your call.

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -277,5 +283,176 @@ pub fn for_print(s: &SharedString) -> SharedString {
...
+                        buf.push(t2);
+                        buf.push(t3);
+                        buf.push(t4);
+
+                        // ...... \uXXXX \uXXXX ......
+                        //        ^      ^      ^
+                        //        |      |      |
+                        //        end    pos    input
+
+                        input = &input[pos + 6..];
```
Yoric: Again this magic `6` :)

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -277,5 +283,176 @@ pub fn for_print(s: &SharedString) -> SharedString {
         };
     }
 
-    SharedString::from_string(String::from_utf8(buf).expect("Escaped string should be valid UTF-8"))
+    String::from_utf8(buf).expect("Unescaped string should be valid UTF-8")
+}
+
+/// If the given `bytes` contains any unicode escape \uXXXX and that is lone
```
Yoric: It's not exactly clear to me what happens if there are multiple escapes, e.g. `"\\\\uXXXX"`. Could you perhaps write this as a doctest?

crates/binjs_io/src/escaped_wtf8.rs
Yoric: (also in `to_unicode_escape`)

crates/binjs_io/src/escaped_wtf8.rs
```diff
@@ -209,14 +217,12 @@ pub fn unescape(bytes: &[u8]) -> Cow<[u8]> {
 }
 
 /// If the given `bytes` contains any escaped lone surropgate, convert it to
-/// \uXXXX format, just for debug print.
-/// This does unrecoverable conversion, the result shouldn't be used outside
-/// of debug print.
+/// unicode escape \uXXXX.
 ///
 /// This assumes the input is well-formed escaped WTF-8, which is the result of
 /// escape function, and panics otherwise.
```
Yoric: Maybe examples/doctests would be useful?

RReverser:
To clarify - this is still for test files, not for Shift <-> BinJS interaction, right? That one still uses internal escaping directly?

arai-a:
This is currently used only in external JSON format (that is input/output for binjs_* commands).

I haven't touched anything around Shift.

PR #267, opened 2019-01-31
Resolve #249 - Entropy 0.3 - the rest

### LINKED ISSUE #249: Grammar extension strategy for entropy
Entropy 0.2, which I'm currently implementing, supports extending floats, strings, etc. but does not support adding interfaces or string enums or even boolean values that do not appear in the dictionary in a specific context.

The following is a simple mechanism for extensibility. It builds upon the existing extension mechanism (prelude / content streams) and does not attempt any kind of optimization, as the idea is that we will update the dictionary every so often to take into account new constructs that have become popular around the web.

# Dictionary generation

When we generate a probability table, we add in every context a single symbol (the "extension value") with the minimal probability.

# Encoding

When we attempt to encode a string enum/interface name/boolean value that does not appear in the probability table in this context, we:
- encode the *extension value*;
- fetch the index of the value from the the prelude dictionary for string enums (respectively interface names, boolean values)
- if there is no index value, append it to the dictionary, creating a new index;
- append the index to the content stream for string enums (respectively interface names, boolean values)

# Decoding

When we extract the *extension symbol* in a context:
- read the index from the content stream for string enums (respectively interface names, boolean values);
- get the value at said index from the prelude dictionary of string enums (respectively interface names, boolean values).

If there is no index or no value, it's a syntax error.

Yoric:
This finishes connecting all the bits and pieces of Entropy 0.3.

With this PR, the compressor can finally compress data whose grammar constructions do not show up in the samples. In particular, this means that we can (poorly) compress without dictionary.

### DIFF
--- a/crates/binjs_shared/src/ast.rs
+++ b/crates/binjs_shared/src/ast.rs
@@ -96,6 +96,15 @@ where
 {
     // Nothing to do.
 }
+impl<I, F> std::fmt::Display for Path<I, F>
+where
+    I: Debug + Display,
+    F: Debug + Display,
+{
+    fn fmt(&self, formatter: &mut std::fmt::Formatter) -> Result {
+        self.items.fmt(formatter)
+    }
+}
 
 #[derive(Debug, PartialEq, Eq, Hash, Clone, Deserialize, Serialize)]
 pub struct PathItem<I, F>
@@ -261,3 +270,4 @@ pub trait Node: downcast_rs::Downcast {
     fn name(&self) -> &'static str;
 }
 impl_downcast!(Node);
+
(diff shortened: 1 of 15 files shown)


crates/binjs_shared/src/ast.rs
```diff
@@ -261,3 +270,4 @@ pub trait Node: downcast_rs::Downcast {
     fn name(&self) -> &'static str;
 }
 impl_downcast!(Node);
+
```
arai-a: remove this maybe?

Yoric: The newline?

arai-a: yes, extra newline at the end of file.
if the diff is shown correctly, now the file ends with `\n\n`, but it should end with `\n`

Yoric: Actually, this gets rid of a missing newline, so I think I'll keep it :)

PR #268, opened 2019-01-31
Moving most of generate_dictionary into the library.

Yoric:
We will use this in followups to script auto-generation of dictionaries during benchmarking.

### DIFF
--- a/src/cli/generate_dictionary.rs
+++ b/src/cli/generate_dictionary.rs
@@ -1,0 +1,127 @@
+//! Generate a static prediction table from a sample of JS source files.
+
+use ::generic::FromJSON;
+use ::io::entropy::dictionary::{Dictionary, DictionaryBuilder, FilesContaining};
+use ::io::statistics::Instances;
+use ::io::{Path as IOPath, TokenSerializer};
+use ::source::{Shift, SourceParser};
+use ::specialized::es6::ast::Walker;
+
+use std::io::{Error, ErrorKind};
+use std::path::Path;
+
+pub struct Options<'a> {
+    pub parser: &'a Shift,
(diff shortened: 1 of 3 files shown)


src/cli/generate_dictionary.rs
```diff
@@ -0,0 +1,127 @@
...
+        return Ok(());
+    }
+
+    progress!(options.quiet, "Parsing.");
+
+    handle_path_or_text(options, shared_builder, shared_number_of_files, source_path)
+}
+
+fn handle_path_or_text<'a>(
+    options: &mut Options<'a>,
```
RReverser: Why does this need to be mutable?

src/cli/generate_dictionary.rs
```diff
@@ -0,0 +1,127 @@
...
+    dictionary_builder: &mut DictionaryBuilder,
+    shared_number_of_files: &mut usize,
```
RReverser: Given that this is always incremented on success, perhaps worth removing a param here and keep track of count in the caller instead?

src/cli/generate_dictionary.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        let sub_dir = sub_dir.join(file_name);
+        for entry in std::fs::read_dir(source_path)?
+            .map(|dir| dir.unwrap())
+        {
+            handle_path(
+                options,
+                shared_builder,
+                shared_number_of_files,
+                &entry.path().as_path(),
```
RReverser: I don't think you need `.as_path()` here - `&PathBuf` already coerces to `&Path`.

Yoric:
Killing this off in favor of #272 .

Issue #270, opened 2019-02-01
`binjs_decode` doesn't initialize properly


dominiccooney:
I expected this to not crash:

````
$ git show --oneline HEAD
8bf15e35 (HEAD -> master, origin/master, origin/HEAD) Merge pull request #266 from arai-a/json-conv-wtf8

$  echo 'function f() {}' | target/release/binjs_encode advanced entropy | RUST_BACKTRACE=1 target/release/binjs_decode advanced entropy 
thread 'main' panicked at 'Cannot get grammar roots "Script"', crates/binjs_io/src/entropy/baseline.rs:194:32
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: std::panicking::begin_panic_fmt
   7: binjs_io::entropy::baseline::BaselineDictionaryBuilder::start::{{closure}}
   8: binjs_io::entropy::baseline::build
   9: binjs_io::entropy::Options::new
  10: <binjs_io::entropy::FormatProvider as binjs_io::FormatProvider>::handle_subcommand
  11: binjs_io::Format::from_matches
  12: binjs_decode::main
  13: std::rt::lang_start::{{closure}}
  14: std::panicking::try::do_call
  15: __rust_maybe_catch_panic
  16: std::rt::lang_start_internal
  17: main
  18: __libc_start_main
  19: _start
````

Yoric:
Apparently, for some reason, the baseline dictionary has difficulties initializing in the decoder (but not in the encoder). It's not the first time we have this kind of issue and we don't catch them because `test_*` only use a single process.

@dominiccooney Is there any chance you could turn your one-liner into a test?

Yoric:
@dominiccooney This should be fixed by 068f02d, if you want to test.

PR #272, opened 2019-02-01
Integration testing between binjs_decode and binjs_encode

Yoric:
This should catch any regression to https://github.com/binast/binjs-ref/issues/270.

If this works, I'll complete this with more tests.

### DIFF
--- a/tests/test_external_roundtrip.rs
+++ b/tests/test_external_roundtrip.rs
@@ -1,0 +1,117 @@
+extern crate env_logger;
+#[macro_use]
+extern crate log;
+extern crate tempdir;
+
+use tempdir::TempDir;
+
+use std::path::Path;
+use std::process::Command;
+
+/// A macro used to simplify debugging.
+const DEBUG: bool = true;
+
+fn run(mut command: std::process::Command) -> Result<(), std::io::Error> {
+    let output = command.output()?;
+    if DEBUG || !output.status.success() {
+        debug!(target: "test", "Command: {command:?}\n** out:{out}\n** err:{err}",
+            command = command,
+            out =  String::from_utf8(output.stdout).unwrap_or_else(|_| "<cannot decode output>".to_string()),
+            err =  String::from_utf8(output.stderr).unwrap_or_else(|_| "<cannot decode output>".to_string())
+        );
+    }
+    assert!(output.status.success());
+    Ok(())
+}
+
+fn perform_test(root: &Path, prefix: &str, source: &str) {
+    let root = root.join(prefix);
+    let path_in_source = std::path::Path::new(source);
+    let path_tmp_source = root.join("sample.js");
+    let path_out_encoded_1 = root.join("encoded_1");
+    let path_in_encoded_1 = root.join("encoded_1").join("sample.binjs");
+    let path_out_encoded_2 = root.join("encoded_2");
+    let path_in_encoded_2 = root.join("encoded_2").join("sample.binjs");
+    let path_decoded = root.join("decoded.js");
+
+    let mut dir_builder = std::fs::DirBuilder::new();
+    dir_builder.recursive(true).create(root).unwrap();
+
+    // Copy source file to `path_tmp_source`, to normalize file name.
+    std::fs::copy(&path_in_source, &path_tmp_source).unwrap();
+
+    debug!(target: "test", "Encoding to encoded_1/sample.binjs");
+    let mut command_1 = Command::new("target/debug/binjs_encode");
+    command_1
+        .arg("--in")
+        .arg(&path_tmp_source)
+        .arg("--out")
+        .arg(&path_out_encoded_1)
+        .args(&["advanced", "entropy"]);
+    run(command_1).unwrap();
+
+    debug!(target: "test", "Decoding to decoded.js");
+    let mut command_2 = Command::new("target/debug/binjs_decode");
+    command_2
+        .arg(&path_in_encoded_1)
+        .arg(&path_decoded)
+        .args(&["advanced", "entropy"]);
+    run(command_2).unwrap();
+
+    // Copy again to `path_tmp_source`, to normalize file name.
+    std::fs::copy(&path_decoded, &path_tmp_source).unwrap();
+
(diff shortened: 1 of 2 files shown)


tests/test_external_roundtrip.sh
```diff
@@ -0,0 +1,6 @@
+# This test ensures that we can call encode | decode as two distinct processes.
+
+echo "external_roundtrip: Baseline dictionary"
+echo 'function f() {}' | target/debug/binjs_encode advanced entropy | target/debug/binjs_decode advanced entropy
```
RReverser: Let's maybe check the output of roundtrip as well? (just for basic sanity check)

Yoric: Yeah, I'm rewriting this to check it.

Yoric:
...and the test fails, at least locally, which seems to indicate a bug.

As far as I can tell, the bug around my use of Brotli. To be continued.

Yoric:
To be more precise, the Brotli decompressor seems to always decompress to 64 bytes, which seems odd.

Yoric:
Ok, actually found two issues:

- one deals with brotli – for some reason, when decompressing Brotli, I need to specify a huge buffer size, otherwise it stops decompressing before the end – this is probably an oddity with the API of Brotli, but it's very, very weird;
- one is clearly my fault, `StringDecoder` didn't understand the null string.

I'll post patches shortly.

Yoric:
Requires #275 before landing.

Yoric:
Mmmh... There was a snafu (my fault) and I accidentally landed this patch already.

RReverser:
Heh... does this mean this PR should be closed?

Yoric:
Well, we might as well have a review of the patch :)

tests/test_external_roundtrip.rs
```diff
@@ -0,0 +1,109 @@
...
+            err =  String::from_utf8(output.stderr).unwrap_or_else(|_| "<cannot decode output>".to_string())
+        );
+    }
+    assert!(output.status.success());
+    Ok(())
+}
+
+fn perform_test(prefix: &str, source: &str) {
+    let path_in_source = std::path::Path::new(source).to_path_buf();
+    let path_tmp_source = get_path(&[prefix, "sample.js"]);
```
RReverser: I think I'd rather use [`tempdir`](https://docs.rs/tempdir) or [`tempfile`](https://docs.rs/tempfile) here which would take care both of creating a directory name that is guaranteed to be unique / not already existing on the filesystem, and also would clean up after themselves rather than leaving a stray directory in the temp folder.

RReverser: Alternatively, with `tempfile` you can just create unique temporary files with the same effect, but not having to do any Path manipulations.

tests/test_external_roundtrip.rs
```diff
@@ -0,0 +1,109 @@
...
+        .unwrap_or_else(|e| panic!("Could not open file {:?}: {:?}", path_in_encoded_1, e));
+    let encoded_2 = std::fs::File::open(&path_in_encoded_2)
+        .unwrap_or_else(|e| panic!("Could not open file {:?}: {:?}", path_in_encoded_2, e));
+    assert_eq!(
+        encoded_1.metadata().unwrap().len(),
+        encoded_2.metadata().unwrap().len(),
+        "Distinct file lengths in test {}",
+        prefix
+    );
+    for (i, (b1, b2)) in encoded_1.bytes().zip(encoded_2.bytes()).enumerate() {
```
RReverser: Reading byte-by-byte, especially on unbuffered `File` (w/o `BufReader`) is quite inefficient and might be potentially noticeable on large test files (either current or ones added in future).

I'd suggest at least reading entire files using `std::fs::read` instead and then comparing the returned vectors.

tests/test_external_roundtrip.rs
```diff
@@ -0,0 +1,116 @@
...
+            command = command,
+            out =  String::from_utf8(output.stdout).unwrap_or_else(|_| "<cannot decode output>".to_string()),
+            err =  String::from_utf8(output.stderr).unwrap_or_else(|_| "<cannot decode output>".to_string())
+        );
+    }
+    assert!(output.status.success());
+    Ok(())
+}
+
+fn perform_test(root: &Path, prefix: &str, source: &str) {
```
RReverser: Hmm, my other comment from prev review seems to got lost.

I suggested to join `root` + `prefix` just once at the top, and then also use `.join` for all files below instead of custom util with array. Something like:

```rust
let root = root.join(prefix);
...
let path_tmp_source = root.join("sample.js");
let path_out_encoded_1 = root.join("encoded_1");
let path_in_encoded_1 = path_out_encoded_1.join("sample.binjs");
...
```

tests/test_external_roundtrip.rs
```diff
@@ -0,0 +1,116 @@
...
+    let path_in_source = std::path::Path::new(source).to_path_buf();
```
RReverser: I don't think you need `to_path_buf` here since `fs::copy` accepts just `&Path` too.

tests/test_external_roundtrip.rs
```diff
@@ -0,0 +1,106 @@
...
+    let path_decoded = root.join("decoded.js");
+
+    let mut dir_builder = std::fs::DirBuilder::new();
+    dir_builder.recursive(true).create(root).unwrap();
+
+    // Copy source file to `path_tmp_source`, to normalize file name.
+    std::fs::copy(&path_in_source, &path_tmp_source).unwrap();
+
+    debug!(target: "test", "Encoding to encoded_1/sample.binjs");
+    let mut command_1 = Command::new("target/debug/binjs_encode");
```
RReverser: Hmm, I just realised that this depends on `binjs_encode` CLI app being already built, but this dependency/order is not enforced it in any way, which means that e.g. `cargo test` on clean checkout would fail.

Not sure if there's a good way to enforce it, but seems like something that should be at least explicitly documented (that you need to run `cargo build` to build all targets before testing)?

Yoric: I have added an explicit call to `cargo build` during the initialization of the test.

PR #282, opened 2019-02-06
Resolve #280 - Force Brutli to flush correctly.

### LINKED ISSUE #280: Brotli doesn't seem to (always?) flush correctly

Yoric:


crates/binjs_io/src/entropy/read/prelude_decoders.rs
```diff
@@ -163,25 +163,14 @@ where
 
         // Extract slice.
         let mut buf = Vec::with_capacity(byte_len); // FIXME: We may want some maximal size here.
-        unsafe {
-            buf.set_len(byte_len);
-        }
+        buf.resize(byte_len, 0);
```
RReverser: Nit: this and the former line can be combined into `vec![0; byte_len]` (as a bonus to clarity, it's also slightly more efficient as it can use `calloc`).

crates/binjs_io/src/entropy/write/lazy_stream.rs
```diff
@@ -24,14 +25,29 @@ const BROTLI_LG_WINDOW_SIZE: u32 = 20;
 /// all data to an external file, for
 /// forensics purposes.
 
+struct Dumps {
+    raw: std::fs::File,
+    brotli: brotli::CompressorWriter<std::fs::File>,
+}
+impl Write for Dumps {
+    fn write(&mut self, data: &[u8]) -> Result<usize, std::io::Error> {
```
RReverser: Suggestion: maybe provide this as an implementation for `write_all` instead (which this essentially is), and then let `write` call into that but return `Ok(data.len())`.

crates/binjs_io/src/entropy/read/prelude_decoders.rs
Yoric: Ooooohhh...
I had never heard of this syntax, thanks!

crates/binjs_io/src/entropy/write/lazy_stream.rs
```diff
@@ -76,67 +95,93 @@ impl LazyStream {
...
-            return Ok(Some(self.lazy_dump_file.as_mut().unwrap()));
+        if let Some(mut path) = self.dump_path.take() {
+            {
+                let dir = path.parent().unwrap();
+                std::fs::DirBuilder::new().recursive(true).create(dir)?;
+            }
+
+            let raw = std::fs::File::create(&path)?;
+
+            let extension = match path.extension() {
```
RReverser: Technically, `.extension()` / `.set_extension()` are meant to be used for a single extension, which is why this becomes a bit verbose. For our case, maybe it would be simpler to use string-based manipulation here?

```rust
let mut path = path.into_os_string();
path.push(".bro");
```

crates/binjs_io/src/entropy/write/lazy_stream.rs
```diff
@@ -24,14 +25,33 @@ const BROTLI_LG_WINDOW_SIZE: u32 = 20;
 /// all data to an external file, for
 /// forensics purposes.
 
+struct Dumps {
+    raw: std::fs::File,
+    brotli: brotli::CompressorWriter<std::fs::File>,
+}
+impl Write for Dumps {
+    fn write_all(&mut self, data: &[u8]) -> Result<(), std::io::Error> {
```
RReverser: As on the other PR, maybe use `std::io::Result` alias for IO implementations?

crates/binjs_io/src/entropy/read/prelude_decoders.rs
```diff
@@ -162,26 +162,14 @@ where
...
-
         let mut result = Vec::new();
-        {
-            let mut brotli = brotli::DecompressorWriter::new(&mut result, 32768); // FIXME: WTF? This doesnt seem to work with small buffer sizes. How am I supposed to guess the buffer size?
-            brotli
-                .write_all(&buf)
-                .map_err(TokenReaderError::ReadError)?;
-            brotli.flush().map_err(TokenReaderError::ReadError)?;
-        }
+        brotli::BrotliDecompress(&mut Cursor::new(buf), &mut result)
```
RReverser: On the second thought... why do we need temporary buffer here? Can we pass slice of input directly to Brotli?

I imagine passing something like `self.input.take(byte_len)` should work (maybe with some casting to avoid consuming `self.input` but rather `&mut self.input` though, or maybe Rust will be smart enough to figure this out itself).

crates/binjs_io/src/entropy/write/lazy_stream.rs
Yoric: I have no strong opinion on this – especially since this code is dedicated to debugging/forensics.

RReverser: I'm mostly suggesting this for readability (the string version seems simpler), but it's up to you.

crates/binjs_io/src/entropy/read/prelude_decoders.rs
Yoric: Yes, we could use `self.input.take`, which would save some copies and allocations, but in case of encoding bug, it's going to make it harder for use to find the issue. So, if you don't mind, I'd rather keep the current implementation.

arai-a: would be nice to add comment about that optimization opportunity.

crates/binjs_io/src/entropy/write/lazy_stream.rs
```diff
@@ -24,14 +25,33 @@ const BROTLI_LG_WINDOW_SIZE: u32 = 20;
 /// all data to an external file, for
 /// forensics purposes.
 
+struct Dumps {
```
arai-a: can you add comment what this is for?

crates/binjs_io/src/entropy/read/prelude_decoders.rs
Yoric: Ok, I have added a workaround that should bring us the best of both worlds.

RReverser: Hmm, I'm not sure how is `BoundedRead` different from what `take` does. Could you add some comments please?

Yoric: > Ah are you saying that `BrotliDecompress` silently ignores extra data in the end of input? If so, that's unfortunate...

I don't want to rely upon this (unspecified) behavior. I want to be able to check that the length is exactly the one that was indicated in the header.

> But maybe a simpler solution would be to provide util that only tracks amount of data read, and then compose that with `.take()` instead of reimplementing `.take()` logic too? I believe it should be a bit simpler.

I already had `Pos` at hand, it seemed simpler to reuse it.

RReverser: > I already had `Pos` at hand, it seemed simpler to reuse it.

What is `Pos`? My suggestion is to change the helper to only track position after each `read`, and not check for EOF etc. (leave that to built-in `take`). That should significantly simplify the logic and allow to reuse it in more contexts.

Yoric: Done.

PR #284, opened 2019-02-07
More external roundtrip tests

Yoric:
This patch adds an external roundtrip test that builds and uses an external dictionary.

### DIFF
--- a/tests/test_external_roundtrip.rs
+++ b/tests/test_external_roundtrip.rs
@@ -5,13 +5,14 @@ extern crate tempdir;
 
 use tempdir::TempDir;
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
 /// A macro used to simplify debugging.
 const DEBUG: bool = true;
 
-fn run(mut command: std::process::Command) -> Result<(), std::io::Error> {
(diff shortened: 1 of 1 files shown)


tests/test_external_roundtrip.rs
```diff
@@ -91,27 +116,56 @@ fn perform_test(root: &Path, prefix: &str, source: &str) {
...
 }
 
-/// Ensure that `binjs_encode` and `binjs_decode` are built.
+/// Generate a dictionary from a source.
+///
+/// Return the path of the file containing the generated dictionary.
+fn init_dictionary(root: &Path, source: &str) -> PathBuf {
+    let root = root.join("dictionary");
+    let mut command = Command::new("target/debug/binjs_generate_prediction_tables");
+    command.arg("--in").arg(source).arg("--out").arg(&root);
```
RReverser: Maybe `.args(&[...])` would be simpler with this number of args?

Yoric: Wouldn't work for `root`, so that would give us one `args` and one `arg`, which I don't really find simpler.

RReverser: Why wouldn't?

RReverser: Ah, type mismatch.

Issue #256, opened 2019-01-18
Investigate mechanisms to reduce the length of indices in content streams


Yoric:
We currently define indices as:

- by default, an index into the shared dictionary;
- if the index is larger than the the number of items in the shared dictionary, an index into a prelude dictionary.

This is possibly wasteful, as there are many items in the shared dictionary that we are bound to never use, so we end up using large varnums where we could use shorter ones. We should consider a mechanism to ensure that we assign short varnums as often as possible.

Yoric:
For some streams, coding the index as a difference from the previous index might work nicely.

dominiccooney:
I have a different idea for this I'd like to try in addition to what you have below. Should I file a sub-issue for it? (To be honest I find the number of issues a bit tough to keep track of, and I can't see how to close issues.)

> # A few ideas
> * sanity checks:
>   
>   * [ ]  when we parse `foo["bar"]`, we should interpret this as `foo.bar` (with a `PropertyKey`, rather than a `LiteralString`), do we really do it? If not, we're polluting the space of literal strings with values that are actually property keys, and we're missing compression opportunities.

This makes toString lower fidelity and harder because it needs to scan identifiers. What guides decisions like that?

Yoric:
> I have a different idea for this I'd like to try in addition to what you have below. Should I file a sub-issue for it?

Ideally, both file a sub-issue and put the issue on the list above.

> (To be honest I find the number of issues a bit tough to keep track of, and I can't see how to close issues.)

For some reason, you were not part of the list of owners. Sorry about that, I guess the list didn't carry over when we renamed the repo. You should now have access to a "Close" button.

> > # A few ideas
> > 
> > * sanity checks:
> >   
> >   * [ ]   when we parse `foo["bar"]`, we should interpret this as `foo.bar` (with a `PropertyKey`, rather than a `LiteralString`), do we really do it? If not, we're polluting the space of literal strings with values that are actually property keys, and we're missing compression opportunities.
> 
> This makes toString lower fidelity and harder because it needs to scan identifiers.

What do you mean "scan identifiers"? To handle property keys such as `"several words separated by whitespaces"`?

> What guides decisions like whether rewrites like this should be part of the compressor? We could probably assume a BinAST-sympathetic minifier exists in future, maybe transformations like this should be part of a separate frontend piece.

You are right that there is an ambiguity there and that this should belong to a different brick of the toolchain. However, we are faced with two issues:

[...]

PR #285, opened 2019-02-07
Issue #256 - Refactor to simplify experiments on indices.

Yoric:
We currently write indices to Brotli as soon as we encounter them. This patch delays encoding until the final flush. This should make it easier for us to experiment with some changes in the encoding, e.g. differential encoding, or Burrows-Wheeler Transform, etc.


crates/binjs_io/src/entropy/write/mod.rs
```diff
@@ -78,6 +78,8 @@ pub struct Encoder {
 
     /// Measure the number of entries written.
     content_instances: ContentInfo<Instances>,
+
+    path: Option<std::path::PathBuf>,
```
arai-a: Add comment maybe?

PR #287, opened 2019-02-08
New command-line tool: investigate_streams.

Yoric:
This tool should help us determine which streams are most expensive in terms of disk usage.

### DIFF
--- a/examples/investigate_streams.rs
+++ b/examples/investigate_streams.rs
@@ -36,6 +36,7 @@ use clap::*;
 use itertools::Itertools;
 
 use std::collections::HashMap;
+use std::path::Path;
 use std::process::Command;
 
 /// The sizes observed for a stream.
@@ -46,6 +47,36 @@ struct Sizes {
 
     /// Brotli-compressed size.
     brotli: u64,
+
+    /// Bzip2-compressed size.
+    bzip: u64,
+}
+
+fn get_brotli_size(path: &Path) -> u64 {
+    let out = Command::new("brotli")
+        .arg("--best")
+        .arg("--keep")
+        .arg("--stdout")
+        .arg(&path)
+        .output()
+        .expect("Error during brotli");
+    assert!(out.status.success());
+    assert!(out.stdout.len() != 0);
+    out.stdout.len() as u64
+}
+
+fn get_bzip2_size(path: &Path) -> u64 {
+    let out = Command::new("bzip2")
+        .arg("--compress")
+        .arg("--best")
+        .arg("--keep")
+        .arg("--stdout")
+        .arg(&path)
+        .output()
+        .expect("Error during bzip2");
+    assert!(out.status.success());
+    assert!(out.stdout.len() != 0);
+    out.stdout.len() as u64
 }
 
 fn main() {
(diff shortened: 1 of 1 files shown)


examples/investigate_streams.rs
```diff
@@ -0,0 +1,266 @@
...
+        "File",
+        &Compressions {
+            raw: "raw (b)",
+            brotli: "brotli (b)",
+            bzip: "bzip2 (b)",
+            lzma: "lzma (b)",
+        },
+    );
+    let print = |in_extension: &HashMap<String, Compressions<u64>>| {
+        for (name, info) in in_extension.into_iter().sorted().into_iter() {
```
RReverser: Maybe instead of `.sorted()` here it's simpler to use `BTreeMap` in the first place? I imagine performance difference is not important here anyway.

examples/investigate_streams.rs
```diff
@@ -0,0 +1,266 @@
...
+            }
+            // Brotli files, output by binjs_encode.
+            // Remove the `bro` from the name, store in `Compressions::brotli`.
+            "bro" => {
+                let path = path.as_path().with_extension("");
+                let name = path.file_name().unwrap().to_str().unwrap().to_string();
+                let extension = path.extension().unwrap().to_str().unwrap().to_string();
+                let info = stats
+                    .entry(extension.clone())
+                    .or_insert_with(|| HashMap::new())
```
RReverser: Nit: you can use `.or_default()` for all these instances.

examples/investigate_streams.rs
Yoric: Especially since we're sorting ~10 items.

examples/investigate_streams.rs
Yoric: Actually, it requires type annotations if I replace everything with `or_default()`. I figure it's not worth the tradeoff, what do you think?

RReverser: It might require one annotation for the `stats` where it's defined, but the call sites should be simpler.

examples/investigate_streams.rs
RReverser: Yeah, so `BTreeMap` should be simpler. (Is that what you meant by resolving this conversation?)

Yoric: Oh, I probably misread your previous comment. Yeah, you're right, let's use `BTreeMap`.

examples/investigate_streams.rs
Yoric: Ok, will do.

examples/investigate_streams.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    // Output `js` and `binjs` first.
+    if let Some(in_extension) = stats.remove("js") {
+        print(&in_extension);
+    }
+    if let Some(in_extension) = stats.remove("binjs") {
+        print(&in_extension);
+    }
+    // Output the rest by alphabetical order.
+    for extension in stats.keys().sorted().into_iter() {
```
RReverser: Maybe `BTreeMap` for top-level map as well?

examples/investigate_streams.rs
```diff
@@ -0,0 +1,268 @@
...
+            raw: "raw (b)",
+            brotli: "brotli (b)",
+            bzip: "bzip2 (b)",
+            lzma: "lzma (b)",
+        },
+    );
+    let print = |in_extension: &BTreeMap<String, Compressions<u64>>| {
+        // Note: we sort here to always have the same traversal order across runs.
+        // Performance doesn't matter all that much, as we're sorting only ~5 times, at most 10 items at once.
+        for (name, info) in in_extension.into_iter() {
```
RReverser: You probably don't need `.into_iter()` anymore.

examples/investigate_streams.rs
RReverser: Also, I think you could instead use

```rust
for (extension, in_extension) in stats {
  print(in_extension);
}
```

after the type change.

PR #301, opened 2019-02-13
Resolve #298 - In content streams, reserve 0 for "next in prelude"

### LINKED ISSUE #298: In content streams, see what happens if we reserve 0 for "next in prelude".
Consider the string content stream. Assuming that most strings are in the prelude and are not repeated, the most common pattern will be "next string in prelude". If we always encode this as `0`, this might make the stream easier to compress.

Yoric:
As it turns out, the most common value we fetch from a content stream is the "next in prelude". Experiments (see issue #298) indicate that reserving 0 for "next in prelude" improves compression of most streams. This is what this patch does.

### DIFF
--- a/crates/binjs_io/src/entropy/dictionary.rs
+++ b/crates/binjs_io/src/entropy/dictionary.rs
@@ -110,99 +110,138 @@ macro_rules! increment_instance_count {
     }
 }
 
-/// Result of a fetch operation in a IndexedTable.
+/// Result of a fetch operation in a LinearTable.
 #[derive(Clone, Copy, Debug)]
 pub enum Fetch {
     /// The value was already in the cache at the given index.
-    Hit(Index),
+    Hit(TableRef),
 
     /// The value was not in the cache. A slot has been allocated at the given index,
     /// but the definition still needs to be added.
-    Miss(Index),
+    Miss(TableRef),
 }
 impl Fetch {
-    /// Return the raw value of the fetch, as a `usize`, for writing to a stream.
-    pub fn raw(&self) -> usize {
+    /// Return `true` if this result represents a hit.
+    pub fn is_hit(&self) -> bool {
         match *self {
-            Fetch::Hit(result) => result.raw(),
-            Fetch::Miss(result) => result.raw(),
+            Fetch::Hit(_) => true,
+            _ => false,
+        }
+    }
+
+    pub fn table_ref(&self) -> TableRef {
+        match *self {
+            Fetch::Hit(result) => result,
+            Fetch::Miss(result) => result,
         }
     }
 }
 
-/// An index in an IndexedTable.
-#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
-pub enum Index {
+/// An index in an LinearTable.
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub enum TableRef {
     /// The index represents a value in a shared dictionary.
     Shared(usize),
 
     /// The index represents a value in a prelude dictionary.
     Prelude(usize),
 }
-impl Index {
-    /// Return the raw value of the index, as a `usize`, for writing to a stream.
-    pub fn raw(&self) -> usize {
+impl TableRef {
+    /// Return the `usize` carried by this `TableRef`.
+    fn raw(&self) -> usize {
         match *self {
-            Index::Shared(result) => result,
-            Index::Prelude(result) => result,
+            TableRef::Shared(raw) => raw,
+            TableRef::Prelude(raw) => raw,
         }
     }
 }
 
-/// The initial size of IndexedTables, in elements.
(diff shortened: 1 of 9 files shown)


crates/binjs_io/src/entropy/read/content_decoders.rs
```diff
@@ -79,8 +85,20 @@ where
                     Ok(result) => result as usize,
                     Err(err) => return Some(Err(TokenReaderError::ReadError(err))),
                 };
-
                 debug!(target: "read", "DictionaryStreamDecoder::next index: {}", index);
+                let index = if index == 0 {
+                    // Special case: 0 is reserved for "next in prelude".
```
arai-a: to me, "next in prelude" isn't clear what it means (especially what the unit for "next" is)
can it be "next entry in prelude dictionary" or something?

crates/binjs_io/src/entropy/write/mod.rs
```diff
@@ -308,9 +308,30 @@ impl Encoder {
 
         // Write (and possibly dump) data.
         // In the current implementation, we just ignore any information other than the index.
+
+        // We reserve number 0 for "next in prelude", as it is a very common pattern, and having
```
arai-a: here too.

crates/binjs_io/src/entropy/read/content_decoders.rs
```diff
@@ -29,6 +29,11 @@ pub struct DictionaryStreamDecoder<'a, T> {
 
     /// The name of this dictionary. Used for debugging/error reporting.
     name: SharedString,
+
+    /// The latest index we have read from the prelude dictionary.
+    ///
+    /// Used to decode the special symbol `0`.
```
arai-a: can you explain what this value is used for?
here, or add reference to other method.

(IMO, it's more important than what the consumer is)

crates/binjs_io/src/entropy/write/mod.rs
```diff
@@ -308,9 +308,30 @@ impl Encoder {
...
+        // successive 0s improves compression (confirmed by issue #298).
+        // Any other index i is written as i + 1.
+        let mut latest_in_prelude = None;
         for item in vec {
-            let index = item.raw();
-            lazy_stream.write_varnum(index as u32)?;
+            let varnum = match *item {
+                // References to the shared dictionary are shifted by 1.
+                Index::Shared(i) => i + 1,
+                // References to the prelude dictionary may be 0 (for next) or shifted by 1.
```
arai-a: Please move the comment and code about the 0/shifting to `Index` enum's declaration/implementation,
and not to embed in random place.
both encoding and decoding.

arai-a:
To be clear, by "error-prone", I mean, the `Index` data can be used in other wrong way with current code, like forgetting shifting when creating instance or getting value.
So, changing the `Index` implementation not to allow such code would be better.

crates/binjs_io/src/entropy/dictionary.rs
```diff
@@ -110,99 +110,138 @@ macro_rules! increment_instance_count {
...
-    /// Create an empty `IndexedTable`.
+    /// Create an empty `LinearTable`.
     pub fn with_capacity(len: usize) -> Self {
-        IndexedTable {
+        LinearTable {
             values: Vec::with_capacity(len),
-            indices: HashMap::with_capacity(len),
+            refs_by_value: HashMap::with_capacity(len),
+            shared_len: 0,
         }
```
arai-a: might be nice to add constructor (new static method?)

crates/binjs_io/src/entropy/dictionary.rs
```diff
@@ -223,11 +262,24 @@ where
...
+        for item in prelude {
+            if clone.fetch_index(item).is_hit() {
+                // The prelude shouldn't duplicate anything from the dictionary.
+                return Err(item);
+            }
+        }
+        Ok(clone)
+    }
+
     /// Return the current state of the cache as a slice.
```
arai-a: this comment seems to be obsolete

crates/binjs_io/src/entropy/read/mod.rs
```diff
@@ -337,13 +357,13 @@ macro_rules! content_stream {
         {
             debug!(target: "read", "Reading from content stream {}", $description);
             let value = $me.$stream.next()
-                .unwrap_or_else(|| Err(TokenReaderError::UnexpectedEndOfStream($description.to_string())))?;
+                .unwrap_or_else(|| if true { panic!(); } else { Err(TokenReaderError::UnexpectedEndOfStream($description.to_string())) } )?;
```
arai-a: revert this maybe?

crates/binjs_io/src/entropy/dictionary.rs
Yoric: There are already two of them a few lines below :)

PR #307, opened 2019-02-15
Resolves #305 - Implement a mechanism to cleanup `foo["bar"]` into `f…

### LINKED ISSUE #305: Implement a mechanism to cleanup foo["bar"] into foo.bar
Since we namespace strings and property keys separately, having both `foo["bar"]` and `foo.bar` in the same source hurts compression. We should have a mechanism to optionally clean this up.

I believe that the best place to do so is around the Shift parser.

Yoric:
…oo.bar`.

We implement this mechanism during Shift conversion. With this patch, the cleanup is enabled by default.

### DIFF
--- a/src/source/shift.rs
+++ b/src/source/shift.rs
@@ -125,11 +125,27 @@ impl Script {
     }
 }
 
+/// Options for parsing or pretty printing.
+struct Options {
+    /// If `true`, rewrite `foo["bar"]` into `foo.bar` whenever possible.
+    ///
+    /// Default value: `true`.
+    cleanup_bracket_expressions: bool,
+}
+impl Options {
+    fn new() -> Self {
+        Options {
+            cleanup_bracket_expressions: true,
+        }
+    }
+}
+
 /// Using a Node + Shift binary to parse an AST.
 pub struct Shift {
     parse_str: Script,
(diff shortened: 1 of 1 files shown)


RReverser:
Hmm how common is that case? Gut feeling teels me that conversion like this should live in a separate tool (like minifier) perhaps as it doesn't have much to do with BinaryAST itself...

Yoric:
I agree that it should, ideally, live in a separate tool. I'm not sure how we could produce such a tool, at the moment, though, at least not with important refactoring. Do you have any idea?

Also, my objective with this patch was to let us test whether this improves compression, but I guess I can perform the tests without landing the patch.

RReverser:
> I'm not sure how we could produce such a tool, at the moment, though, at least not with important refactoring.

I think this is already a feature of most popular minifiers, and I'd expect most popular assets to be minified regardless of BinaryAST.

E.g. with UglifyJS:

```
> echo 'x["bar"]' | uglifyjs --compress
x.bar;
```

Yoric:
Yeah, I was thinking of the cases that cannot be represented easily within the JS syntax (e.g. `foo["several words & special characters"]`), but they are probably not very common :)

Let's forget about this PR for now.

Issue #296, opened 2019-02-12
Embedded sublanguages


Yoric:
JavaScript has subsets that we can probably encode more efficiently, either with different dictionaries or different strategies.

For instance, if we know by earlier analysis that an expression is a JSON-style literal declaration, we could shift, for this subtree, to a dictionary that assigns non-0 probabilities only to literals. This would make compression more efficient on that subtree.

Similarly, if we know by earlier analysis that an expression is an arithmetic expression, we could shift to an arithmetic expression dictionary.

Yoric:
One way to do this would be to:

1/ somehow patch our webidl to expand `Expression` and add `@@LiteralExpression`, `@@ArithmeticExpression`;
2/ have Shift=>ast conversion analyze the values it converts and, if useful, produce instances of `@@LiteralExpression` or `@@ArithmeticExpression` instead of the more general variants;
3/ similarly, convert back from `@@LiteralExpression` or `@@ArithmeticExpression` into their respective general variants when converting ast=>Shift.

Yoric:
Perhaps we could somehow inject a node

```webidl
interface BinASTExpressionWithProbabilityTable {
   expression: Expression;
   table: string;
}
```

into `Expression`, and then rely on heuristics to decide when to replace an expression `foo` with `BinASTExpressionWithProbabilityTable { foo, table }`.

Yoric:
For the time being, I'm going to assume that we can switch the probability tables, but not the content dictionaries.

Yoric:
On the Facebook test set, I see a very, very slight improvement (less than 1/1000).

Yoric:
Testing with https://github.com/binast/binjs-ref/pull/372

# Data


File | raw | brotli |  ratio vs. master
-- | -- | -- | --
js | 43134534 | 8016723 | 1
binjs | 8930118 | 8860979 | 1.00058085996408
  |   |   |  
floats.content | 336910 | 110130 | 1
floats.prelude | 126094 | 72487 | 1
identifier_names.content | 2524124 | 997583 | 1
identifier_names.prelude | 86304 | 52927 | 1
identifier_names_len.prelude | 26637 | 15806 | 1
list_lengths.content | 1985830 | 550497 | 1
list_lengths.prelude | 10284 | 12132 | 1
main.entropy | 1598083 | 1590968 | 0.993700399861591
property_keys.content | 1630992 | 986205 | 1
property_keys.prelude | 3150015 | 1010811 | 1
[...]

PR #309, opened 2019-02-18
Issue #296 - Extending the webidl with a secondary file.

Yoric:
This patch adds support for extending es6.webidl with an additional file es6-extension.webidl. The role of that file is to modify the AST in a manner that is useful for *compression* but has no impact on visible *syntax and semantics* of the language. In other words, es6.webidl is meant to be part of the BinAST language specification, while es6-extension.webidl is meant to be part of the compression format.

The main objective of this patch is to let us adopt specialized compression dictionaries for e.g. JSON, arithmetic expressions, etc.

### DIFF
--- a/crates/binjs_es6/build.rs
+++ b/crates/binjs_es6/build.rs
@@ -9,23 +9,33 @@ use std::env;
 use std::fs::*;
 use std::io::*;
 
-const PATH_GRAMMAR_ES6: &'static str = "../../spec/es6.webidl";
+/// The webidl source files.
+///
+/// Order is important: first the main file, then any extension.
+const PATH_SOURCES: [&'static str; 2] =
+    ["../../spec/es6.webidl", "../../spec/es6-extension.webidl"];
 
 fn main() {
-    println!("cargo:rerun-if-changed={}", PATH_GRAMMAR_ES6);
+    for source in &PATH_SOURCES {
+        println!("cargo:rerun-if-changed={}", source);
+    }
 
     // Load webidl.
 
-    let mut file = File::open(PATH_GRAMMAR_ES6).expect("Could not open source");
-    let mut source = String::new();
-    file.read_to_string(&mut source)
-        .expect("Could not read source");
+    let sources: Vec<_> = PATH_SOURCES
+        .iter()
+        .map(|path| {
+            read_to_string(path)
+                .unwrap_or_else(|e| panic!("Could not open grammar file {}: {}", path, e))
+        })
+        .collect();
+    let sources = sources.iter().map(String::as_str);
 
     // Check spec. We don't really need fake_root
     // for this operation. It may change in the future,
     // we'll see then.
 
-    let mut builder = Importer::import(&source).expect("Could not parse source");
+    let mut builder = Importer::import(sources).expect("Could not parse webidl sources");
     let fake_root = builder.node_name("@@ROOT@@"); // Ignored.
     let null = builder.node_name(""); // Actually used
     builder.add_interface(&null).unwrap();
--- a/crates/binjs_meta/src/import.rs
+++ b/crates/binjs_meta/src/import.rs
@@ -21,11 +21,15 @@ pub struct Importer {
 impl Importer {
     /// Import a WebIDL spec into a SpecBuilder.
     ///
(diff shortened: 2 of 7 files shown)


spec/es6-extension.webidl
```diff
@@ -0,0 +1,13 @@
...
+
+[ExtendsTypeSum=Expression]
+// An extension to `Expression` used to locally change probability tables.
+interface BinASTExpressionWithProbabilityTable {
+    // The actual expression.
+    attribute Expression expression;
+
+    // The name of the probability table to use (e.g. `main`, `json`).
+    attribute string table;
+};
```
arai-a: missing newline

spec/es6-extension.webidl
```diff
@@ -0,0 +1,13 @@
+// Extensions to the webidl.
+//
+// These extensions are not part of the AST spec, but of the compression mechanism.
+
+[ExtendsTypeSum=Expression]
+// An extension to `Expression` used to locally change probability tables.
+interface BinASTExpressionWithProbabilityTable {
```
arai-a: Can you add some more comment about the purpose of this, and how it's used?
also, will this be exposed to decoder?

crates/binjs_meta/src/import.rs
```diff
@@ -69,13 +73,15 @@ impl Importer {
     ///     assert_eq!(contents_field.is_lazy(), true);
     /// }
     /// ```
-    pub fn import(s: &str) -> Result<SpecBuilder, weedle::Err<CompleteStr>> {
+    pub fn import<'a>(sources: &'a [&str]) -> Result<SpecBuilder, weedle::Err<CompleteStr<'a>>> {
```
RReverser: Perhaps better to use `sources: impl IntoIterator<Item = &str>` to support arbitrary iterables.

Yoric: Good point.

spec/es6-extension.webidl
Yoric: Ok, attempting to clarify.

In practice, the decoder will need to handle decompression of `BinASTExpressionWithProbabilityTable`. This should hopefully be a trivial rule.

crates/binjs_es6/build.rs
```diff
@@ -9,23 +9,33 @@ use std::env;
...
+        println!("cargo:rerun-if-changed={}", source);
+    }
 
     // Load webidl.
 
-    let mut file = File::open(PATH_GRAMMAR_ES6).expect("Could not open source");
-    let mut source = String::new();
-    file.read_to_string(&mut source)
-        .expect("Could not read source");
+    let sources: Vec<_> = PATH_SOURCES
```
RReverser: Now that iterators are supported, I don't think you need intermediate vectors anymore?

RReverser: Ah... you might need them for the reference to point somewhere. Could get away with more generics, but probably not worth it at this point.

spec/es6-extension.webidl
```diff
@@ -0,0 +1,27 @@
...
+    // The list of probability tables is not specified as part of the grammar.
+    // Rather, individual compression dictionaries may define separate probability
+    // tables.
+    //
+    // Attempting to (de)compress with a probability table that does not
+    // appear in the dictionary is a (de)compression error.
+    attribute string table;
+
+    // The actual expression.
+    attribute Expression expression;
```
arai-a: does this `Expression` contain `BinASTExpressionWithProbabilityTable` ?
how is the extension supposed to work regarding recursiveness?

spec/es6-extension.webidl
arai-a: So, it's not implicitly resolved inside decompression, but exposed to token stream, right?

Yoric: Yes. Maybe we'll find a better way to do so in the future, but right now, the idea is to experiment with it.

spec/es6-extension.webidl
Yoric: Yes, this `Expression` may contain ``BinASTExpressionWithProbabilityTable`.

You could switch to a sublanguage inside a sublanguage. For the moment, I have no problem with that.

arai-a:
r+ with those detail about the experiment explained in the comment there.

PR #310, opened 2019-02-18
Introduce a LRU cache for the content identifier stream

Yoric:


crates/binjs_io/src/entropy/read/content_decoders.rs
```diff
@@ -39,23 +39,41 @@ where
...
+                    indexed_dictionary,
+                    stream: None,
+                    name,
+                    index_stream_state: TableRefStreamState::new(0), // Unused
+                })
+            }
+            Some(stream) => {
+                let mut stream = Cursor::new(stream);
+                let window_len =
+                    stream.read_varnum().map_err(TokenReaderError::ReadError)? as usize;
```
arai-a: is this format explained somewhere?

crates/binjs_io/src/entropy/rw.rs
```diff
@@ -161,8 +161,9 @@ pub type NameData = SmallVec<[u8; NAME_MAX_LEN]>;
 ///
 /// `TableRef` values are represented as follows:
 /// - a value `0` represents an `TableRef::Prelude` with index 1 + the latest `TableRef::Prelude` value we have encoded/decoded;
-/// - a value `v` in `[1, length of the shared dictionary]` represent `TableRef::Shared` with index `v - 1`;
-/// - a value `v > length of the shared dictionary` represents `TableRef::Prelude` with index `v - 1`.
+/// - a value `v` in `[1, window len]` represents a repeat of a previous value: 1 it the latest value, 2 the second latest, etc.
+/// - a value `v` in `[window len + 1, window len + length of the shared dictionary]` represent `TableRef::Shared` with index `v - window len - 1`;
+/// - a value `v > window len + length of the shared dictionary` represents `TableRef::Prelude` with index `v - window len - 1`.
```
arai-a: why do we use `v - window len - 1` instead of `v - window len - length of the shared dictionary - 1` ?
can you explain it somewhere?  (otherwise this now looks wrong copy-paste)

crates/binjs_io/src/entropy/rw.rs
```diff
@@ -174,16 +175,37 @@ pub struct TableRefStreamState<T> {
...
     /// Create a new `TableRefStreamState`.
-    pub fn new() -> Self {
+    ///
+    /// `window_max_len` represents the number of integers to reserve to
+    /// represent backreferences to values encountered recently.
+    pub fn new(window_max_len: usize) -> Self {
         TableRefStreamState {
             latest_in_prelude: None,
             phantom: PhantomData,
+            window_max_len,
```
RReverser: Do we need this as a separate field or is it essentially same as `window.capacity()`?

tests/test_dictionary_builder.rs
```diff
@@ -254,16 +254,29 @@ test!(test_linear_table, {
     let indices = (1..10)
         .map(|x| x * 2)
         .chain((1..9).map(|x| x * 2 + 1))
+        .chain((7..9).map(|x| x * 2 + 1))
```
RReverser: Some comments here would be nice IMO.

crates/binjs_io/src/entropy/read/content_decoders.rs
Yoric: It's now a little bit documented in write/mod.rs.

Once the format stabilizes, I'll need to make a serious pass of format documentation.

crates/binjs_io/src/entropy/rw.rs
Yoric: It's essentially the same as `window.capacity()`, but keeping them separate lets me assert that I'm not doing something stupid that will accidentally extend `window.capacity()`.

Do you think I should hide this behind a `#[cfg(debug_assertions)]`?

crates/binjs_io/src/entropy/rw.rs
Yoric: Wrong copy-paste, you're right.

crates/binjs_io/src/entropy/rw.rs
RReverser: Hmm fair enough. I think this is fine as-is then.

crates/binjs_io/src/entropy/rw.rs
```diff
@@ -198,41 +220,116 @@ where
...
+                let next_in_prelude = self.next_in_prelude();
+                break TableRef::Prelude(next_in_prelude as usize);
+            }
+            // Renormalize from 0.
+            let value = value - 1;
+            debug!(target: "rw", "TableRefStreamState::from_u32, value is not 0, renormalizing to {}",
+                value);
+            if (value as usize) < self.window_max_len {
+                // this is a hit in the floating window.
+                break self.window.get(value as usize)?.clone();
```
RReverser: `TableRef` implements `Copy` so `.clone()` shouldn't be necessary?

PR #317, opened 2019-02-20
Resolves #313 - Get rid of hardcoded path /tmp

### LINKED ISSUE #313: Fix usages of /tmp
Bunch of tests in `binjs_io` as well as `compare_compression` example rely on `/tmp/...` for temporary files.

This is not cross-platform and should use `tempdir` crate which we already use in new test code.

Yoric:
As requested!


RReverser:
Thanks!

PR #319, opened 2019-02-20
Remove lifetime from Serialization trait

RReverser:
In all cases `Serialization` already needs only a borrowed value, similarly to `serde::Serialize`, or an owned value of a primitive, which is easy and cheap to get from a borrow.

Hence, the definition can get simpler by always assuming a borrowed value of any type instead of requiring implementations with an explicit lifetime param.

### DIFF
--- a/crates/binjs_es6/src/io.rs
+++ b/crates/binjs_es6/src/io.rs
@@ -33,7 +33,7 @@ where
     }
 }
 
-impl<R> Deserialization<R, Option<bool>> for Deserializer<R>
+impl<R> Deserialization<Option<bool>> for Deserializer<R>
 where
     R: TokenReader,
 {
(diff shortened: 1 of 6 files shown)


RReverser:
Update: removed unused type params from `Serialization` / `Deserialization` traits too, which makes type bounds (e.g. in `encode` / `decode`) much simpler.

RReverser:
Idk, was just looking through generated code and then at the definitions to see if it could get any simpler.

Particularly it bothered me because Rust seems to spend ~30 sec to rebuild the library tree on any change in `binjs_meta` (even a whitespace change) which translates into a poor RLS experience, and while this particular change didn't help / I still don't know the cause, I thought it's still useful to simplify what's possible.

PR #321, opened 2019-02-21
Add --gzip switch to allow benchmarking size against gzip

vdjeric:
I squashed commits in the sizes.py pull request so my fork only appears to be 14 commits ahead of parent instead of only 1


Yoric:
I'm not sure what you mean, but you are going to squash these patches into a single one, right?

vdjeric:
I was referring to this pull request showing "vdjeric wants to merge 14 commits into binast:master from vdjeric:master" when in reality my fork is only 1 commit ahead of binast:master.

Yoric:
Well, if you look at `git log`, it's really 14 commits long. Maybe you forgot to `git pull --rebase`?

PR #325, opened 2019-02-22
Resolve #304 - Reduce sparsity of shared indices

### LINKED ISSUE #304: Reduce sparsity of shared indices
For the moment, we represent references to user-extensible dictionaries as follows:

- values `[0,  length of the shared dictionary[` are reserved for references to the shared dictionary;
- values `[length of the shared dictionary, ...[` are references to the prelude dictionary.

The first segment is very sparse, while the second is fully dense. We should find a way to make both segments dense, to improve compression.

Yoric:
This applies the transformation presented in #304 to reduce the sparsity of shared indices. It improves compression by ~4.5%.

### DIFF
--- a/crates/binjs_io/src/entropy/dictionary.rs
+++ b/crates/binjs_io/src/entropy/dictionary.rs
@@ -116,15 +116,14 @@ macro_rules! increment_instance_count {
 
 /// Result of a fetch operation in a LinearTable.
 #[derive(Clone, Copy, Debug)]
-pub enum Fetch {
-    /// The value was already in the cache at the given index.
-    Hit(TableRef),
+pub enum Fetch<T> {
+    /// The value was already in the cache.
+    Hit(T),
 
-    /// The value was not in the cache. A slot has been allocated at the given index,
-    /// but the definition still needs to be added.
-    Miss(TableRef),
+    /// The value was not in the cache.
+    Miss(T),
 }
-impl Fetch {
+impl<T> Fetch<T> {
     /// Return `true` if this result represents a hit.
     pub fn is_hit(&self) -> bool {
         match *self {
@@ -133,16 +132,17 @@ impl Fetch {
         }
     }
 
-    pub fn table_ref(&self) -> TableRef {
+    /// Return the slot represented by this fetch.
+    pub fn slot(&self) -> &T {
         match *self {
-            Fetch::Hit(result) => result,
-            Fetch::Miss(result) => result,
+            Fetch::Hit(ref result) => result,
+            Fetch::Miss(ref result) => result,
         }
     }
 }
 
 /// An index in an LinearTable.
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
 pub enum TableRef {
     /// The index represents a value in a shared dictionary.
     Shared(usize),
@@ -151,11 +151,10 @@ pub enum TableRef {
     Prelude(usize),
 }
 impl TableRef {
-    /// Return the `usize` carried by this `TableRef`.
-    fn raw(&self) -> usize {
+    pub fn as_shared(&self) -> Option<usize> {
         match *self {
-            TableRef::Shared(raw) => raw,
-            TableRef::Prelude(raw) => raw,
+            TableRef::Shared(result) => Some(result),
+            _ => None,
         }
     }
 }
@@ -250,14 +249,17 @@ where
     /// If the value is already in the cache, return `Fetch::Hit(index)`, where `index` is the
     /// immutable index of the value. Otherwise, allocate a new slot `index` and return
     /// `Fetch::Miss(index)`.
-    pub fn fetch_index(&mut self, value: &T) -> Fetch {
+    ///
+    /// In both `Fetch::Hit(index)` and `Fetch::Miss(index)`, `0` is the first value, `1` the second
+    /// value, etc.
+    pub fn fetch_index(&mut self, value: &T) -> Fetch<TableRef> {
         use std::collections::hash_map::Entry::*;
         debug!(target: "dictionary", "Dictionary: 'I'm looking for {:?} in {:?}", value, self);
-        let len = self.values.len();
+        let index = self.values.len() - self.shared_len;
         let result = match self.refs_by_value.entry(value.clone()) {
             Occupied(slot) => return Fetch::Hit(*slot.get()),
             Vacant(slot) => {
-                let result = TableRef::Prelude(len);
+                let result = TableRef::Prelude(index);
                 slot.insert(result.clone());
                 result
             }
@@ -268,6 +270,7 @@ where
 
     /// Create a copy of the current LinearTable with an added prelude dictionary.
     pub fn with_prelude<'a>(&self, prelude: &'a [T]) -> Result<Self, &'a T> {
+        debug!(target: "dictionary", "LinearTable with {} shared, {} prelude", self.shared_len(), prelude.len());
         let mut clone = self.clone();
         for item in prelude {
             if clone.fetch_index(item).is_hit() {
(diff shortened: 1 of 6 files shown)


crates/binjs_io/src/entropy/dictionary.rs
```diff
@@ -112,15 +112,14 @@ macro_rules! increment_instance_count {
 
 /// Result of a fetch operation in a LinearTable.
 #[derive(Clone, Copy, Debug)]
-pub enum Fetch {
-    /// The value was already in the cache at the given index.
-    Hit(TableRef),
+pub enum Fetch<T> {
+    /// The value was already in the cache at the given slot.
```
arai-a: please add comment that explains what "slot" is, or reference to where it's explained (like, "see SOMETHING comment")

crates/binjs_io/src/entropy/dictionary.rs
```diff
@@ -246,14 +245,17 @@ where
     /// If the value is already in the cache, return `Fetch::Hit(index)`, where `index` is the
     /// immutable index of the value. Otherwise, allocate a new slot `index` and return
     /// `Fetch::Miss(index)`.
-    pub fn fetch_index(&mut self, value: &T) -> Fetch {
+    ///
+    /// In both `Fetch::Hit(index)` and `Fetch::Miss(index)`, `0` is the first value, `1` the second
```
arai-a: s/index/slot/, or document what the relation between index and slot.

crates/binjs_io/src/entropy/dictionary.rs
```diff
@@ -246,14 +245,17 @@ where
...
+    /// value, etc. Indices returned with `Fetch::Hit` and `Fetch::Miss` will tend to overlap.
```
arai-a: can you explain some more about "overlap" ?
(otherwise this comment is just confusing)

crates/binjs_io/src/entropy/rw.rs
```diff
@@ -161,9 +163,12 @@ pub type NameData = SmallVec<[u8; NAME_MAX_LEN]>;
...
-/// - a value `v` in `[window len + 1, window len + length of the shared dictionary]` represent `TableRef::Shared` with index `v - window len - 1`;
-/// - a value `v > window len + length of the shared dictionary` represents `TableRef::Prelude` with index `v - window len - length of shared dictionary - 1`.
+/// - a value `v` in `[1, window len]` represents a repeat of a previous value: 1 for the latest value, 2 the second latest, etc.
+/// - a value `v` in `[window len + 1, window len + prelude len]` represents a `TableRef::Prelude` with index `v - window len - 1`;
+/// - a value `v` in `[window len + prelude len + 1, window len + prelude len + number of distinct TableRef::Shared encountered so far]`
+///       represents the `v - prelude len - window len`th `TableRef::Shared` encountered by this `TableRefStreamState`;
+/// - a value `v > window len + prelude len + number of distinct TableRef::Shared encountered so far` represents a fresh `TableRef::Shared`
+///       with index `v - window len + prelude len + number of distinct TableRef::Shared encountered so far`.
```
arai-a: can these comments be wrapped into 80 chars or so?
(that would make review easier, given I don't have to scroll horizontally for each line :P )

crates/binjs_io/src/entropy/rw.rs
```diff
@@ -191,6 +196,27 @@ pub struct TableRefStreamState<T> {
     /// integers to reserve to represent backreferences to values encountered
     /// recently.
     window_max_len: usize,
+
+    /// A mapping from shared references (`TableRef::Shared`) to numbers, such
```
arai-a: what is the number?

crates/binjs_io/src/entropy/rw.rs
```diff
@@ -297,21 +335,62 @@ where
...
     }
 
+    /// Mark a `TableRef::Shared` as encountered if necessary, return the information
+    /// necessary to encode it as `u32`.
+    ///
+    /// If this is the first instance of `value` to be encoded, return `Fetch::Miss(index)`,
+    /// where `index` is in `[number of distinct instances of `TableRef::Shared` encountered
+    /// so far, number of distinct instances of `TableRef::Shared` encountered
+    /// so far + size of the shared dictionary[`. Record `number of distinct instances of
+    /// `TableRef::Shared` encountered so far` as the future index for `value`.
```
arai-a: can you use something else than backquote for nested quote?
also, can you define some short-name variable, instead of repeating long description of number each time?
here and other places, so that it's easier to follow what the comment says.

Yoric:
Note: The test failure seems to be due to an error in the test itself. Investigating.

Yoric:
Oh, I've located the bug. It was a bug in a previous patch, I'll see the best way to fix it.

crates/binjs_io/src/entropy/dictionary.rs
```diff
@@ -250,14 +249,18 @@ where
     /// If the value is already in the cache, return `Fetch::Hit(index)`, where `index` is the
     /// immutable index of the value. Otherwise, allocate a new slot `index` and return
     /// `Fetch::Miss(index)`.
-    pub fn fetch_index(&mut self, value: &T) -> Fetch {
+    ///
+    /// In both `Fetch::Hit(index)` and `Fetch::Miss(index)`, `0` is the first value, `1` the second
+    /// value, etc. In particular, you received `Fetch::Hit(i)` and `Fetch::Miss(j)`, `i` MAY BE
+    /// equal to `j`.
```
arai-a: I'm still not sure what this sentence actually means.
does it mean that second fetch to the same thing could be `Hit` while the first fetch returned `Miss`,
or something else?

crates/binjs_io/src/entropy/rw.rs
```diff
@@ -160,10 +162,29 @@ pub type NameData = SmallVec<[u8; NAME_MAX_LEN]>;
...
+///     of distinct references to the shared dictionary encountered *so far*,
+///     represents a `TableRef::Shared` that has already been encountered at
+///     least once in this stream
+///       - window len + prelude len + 1 if it was the first `TableRef::Shared`
+///         encountered in this stream;
+///       - window len + prelude len + 2 if it was the second;
+///       - ...
+/// - a value `v > window len + prelude len + import len` represents the first
+///     instance of a `TableRef::Shared`. Its index in the shared dictionary
+///     is `v - (window len + prelude len + import len)`.
```
arai-a: I feel, describing in the opposite way (like `TableRef:Prelude(0)` maps to `window_len + 1`) is better

crates/binjs_io/src/entropy/rw.rs
```diff
@@ -191,6 +212,34 @@ pub struct TableRefStreamState<T> {
     /// integers to reserve to represent backreferences to values encountered
     /// recently.
     window_max_len: usize,
+
+    /// A mapping from shared references (`TableRef::Shared(s)`) to the number
+    /// of shared references in this stream encountered before `s`. In other
+    /// words, the the first entry added (by chronological order) maps to `0`,
+    /// the second maps to `1`, etc. Used to ensure that `TableRef::Shared` references
+    /// are generally small enough to fit in small and successive varnums.
```
arai-a: what do you mean by "small" ?
what is small?

crates/binjs_io/src/entropy/dictionary.rs
Yoric: It means that we can have a call that returns `Fetch::Hit(0)` and another one that returns `Fetch::Miss(0)`.

arai-a: what's the situation?
how is it expected or unexpected?

crates/binjs_io/src/entropy/rw.rs
Yoric: There are now two descriptions/specifications, one for compression and one for decompression.

It's going to be lots of fun to update whenever I make further changes to the format :)

crates/binjs_io/src/entropy/dictionary.rs
Yoric: I'll just remove this comment. It's not really useful.

crates/binjs_io/src/entropy/rw.rs
```diff
@@ -191,6 +233,35 @@ pub struct TableRefStreamState<T> {
     /// integers to reserve to represent backreferences to values encountered
     /// recently.
     window_max_len: usize,
+
+    /// A mapping from shared references (`TableRef::Shared(s)`) to the number
+    /// of shared references in this stream encountered before `s`. In other
+    /// words, the the first entry added (by chronological order) maps to `0`,
+    /// the second maps to `1`, etc. Used to ensure that `TableRef::Shared` references
+    /// fit in successive values, and are generally small enough to be represented by
+    /// a varnum of 1 or 2 bytes, instead of being all over the place.
```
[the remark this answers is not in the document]
Yoric: All of this should pretty much be written in the commit message. I'll try and replicate some of the information here.

PR #327, opened 2019-02-22
Add the ability to customize window-len-* from the command-line

Yoric:


crates/binjs_io/src/entropy/mod.rs
```diff
@@ -85,6 +85,25 @@ const DEFAULT_WINDOW_LEN_IDENTIFIER_NAMES: usize = 8;
 const DEFAULT_WINDOW_LEN_STRING_LITERALS: usize = 0;
 const DEFAULT_WINDOW_LEN_LIST_LENGTHS: usize = 0;
 
+// Poor man's hack to convert a `const usize` into a `&'static str`.
+macro_rules! stringify_const {
```
RReverser: Can we instead make a macro that would wrap the constants? I imagine it would be more reliable.

Yoric: How so?

RReverser: With something like:

```rust
macro_rules! clap_defaults {
    ($(const $name:ident: $ty:ty = $init:expr;)*) => {
        $(const $name: $ty = $init;)*
        
        mod clap_defaults {
            $(pub const $name: &str = stringify!($init);)*
        }
    };
}

clap_defaults! {
    const DEFAULT_WINDOW_LEN_FLOATS: usize = 0;
    const DEFAULT_WINDOW_LEN_UNSIGNED_LONGS: usize = 0;
    const DEFAULT_WINDOW_LEN_PROPERTY_KEYS: usize = 0;
    const DEFAULT_WINDOW_LEN_IDENTIFIER_NAMES: usize = 8;
    const DEFAULT_WINDOW_LEN_STRING_LITERALS: usize = 0;
    const DEFAULT_WINDOW_LEN_LIST_LENGTHS: usize = 0;
}
```

It should reliably work for constants with different types and more than one digit too, should that change in future.

Yoric: Yeah, that looks nicer. I'll do that.

Yoric: I don't have the reflex of using macros as syntax extensions in Rust yet. Which is strange, since I used to write Camlp4 domain-specific plug-ins all the time :)

PR #332, opened 2019-02-26
Add thread stack size wrapper for binjs_decode

RReverser:
Without this, binjs_decode was failing with stack overflow on large-ish inputs.


RReverser:
>  In the future, we may wish to accept an optional env variable instead of of the hardcoded constant.

Note that Rust already has a built-in one - `RUST_MIN_STACK` - which applies to all threads as a minimum number of bytes.

We could just ditch our `.stack_size(...)` lines altogether in favour of that one, but I suppose it's better to have working defaults.

PR #334, opened 2019-02-26
Adding and calling {TokenReader, TokenWriter}::{enter_scoped_dictiona…

Yoric:
…ry_at, exit_scoped_dictionary_at}

This patch adds new trait methods {TokenReader, TokenWriter}::{enter_scoped_dictionary_at, exit_scoped_dictionary_at} and extends the strongly-typed AST's (de)serialization code to call the methods whenever an AST node instructs
the {TokenReader, TokenWriter} to switch dictionaries.

### DIFF
--- a/crates/binjs_generate_library/src/lib.rs
+++ b/crates/binjs_generate_library/src/lib.rs
@@ -5,6 +5,7 @@ use binjs_meta::export::{TypeDeanonymizer, TypeName};
 use binjs_meta::spec::*;
 use binjs_meta::util::*;
 
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::rc::Rc;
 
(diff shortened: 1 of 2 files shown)


crates/binjs_generate_library/src/lib.rs
```diff
@@ -1110,9 +1112,11 @@ impl<R> InnerDeserialization<{rust_name}> for Deserializer<R> where R: TokenRead
         print_file_structure!(self.reader, \"{name} {{{{\");
 {fields_def}
         print_file_structure!(self.reader, \"}}}}\");
-        Ok({rust_name} {{
+{maybe_exit_scoped_dictionary}
+        let result = Ok({rust_name} {{
 {fields_use}
-        }})
+        }});
+        result
```
arai-a: is there any reason to assign to local variable here?

Yoric: Ah no, it's a leftover from a previous iteration. I'll remove that bit.

Issue #335, opened 2019-02-26
This name is both non-const-lexical-bound and var-bound


Yoric:
With the latest master, attempt to parse the following example:

```js
function f(t, e) {
}
if (foo) // NO braces here.
    function f(t, e) {
    }
```

This fails with

```
This name is both non-const-lexical-bound and var-bound: IdentifierName(Dynamic("f"))', crates/binjs_es6/src/scopes.rs:251:13
```

Yoric:
@arai-a you have touched this code more recently than me, do you want to take a look?

arai-a:
do you mean this is a regression?

Yoric:
No, I mean that you may understand that piece of code better than I do these days :)

If I'm wrong, don't worry, I'll take a look at it this week.

arai-a:
it's this case
https://tc39.github.io/ecma262/#sec-functiondeclarations-in-ifstatement-statement-clauses
and I think we don't handle Annex B things.

Yoric:
See, I knew you knew that stuff better than me :)

Maybe we should find a way to fail gracefully instead of asserting?

arai-a:
will try adding the handling, at least for strict mode.

arai-a:
to properly handle non-strict-mode code, we need to add block statement, or something like that (maybe, new interface for implicit block?), in order to store scope information for the function.
that needs to be handled in the spec I think.
I'll address strict mode case here.

RReverser:
> we need to add block statement

Could you elaborate please? Where would it be inserted?

In non-strict code the second function is meant to shadow the first one if the condition is true, so they should share the same scope.

arai-a:
I misunderstood the current situation.
currently we're adding block with lexical scope for `if (foo) { function f() {} }`, but apparently that's wrong. (I was trying to align `if (foo) function f() {}` with that)

so, this should be fixed by changing the function declaration not-to-use lexical scope.

PR #339, opened 2019-02-27
Check function declaration in then/else of if statement and throw error

arai-a:
partial fix for https://github.com/binast/binjs-ref/issues/335

I think, we should add implicit block statement on non-strict mode, but I'd like to wait for @syg's response.
so far, changed it to throw both on strict and non-strict mode, while converting Shift AST to BinAST tree.

and also modified scope handling to throw better panic message when unexpected AST comes.

### DIFF
--- a/src/source/shift.rs
+++ b/src/source/shift.rs
@@ -172,7 +172,7 @@ impl SourceParser for Shift {
 
     fn parse_str(&self, data: &str) -> Result<JSON, Error> {
         let mut ast = self.parse_str.transform(&data.into())?;
-        FromShift.convert(&mut ast);
+        FromShift.convert(&mut ast)?;
         Ok(ast)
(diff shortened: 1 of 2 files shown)


src/source/shift.rs
```diff
@@ -485,8 +486,39 @@ impl FromShift {
             Some("IdentifierExpression") => {
                 debug!(target: "Shift", "FromShift IdentifierExpression {:?}", object);
             }
+            Some("IfStatement") => {
```
RReverser: Hmm could this check be moved to use the resulting typed format? I'm trying to get rid of an intermediate JSON AST, and, in particular, started with moving these conversions in a parallel PR.

arai-a:
turns out this fix is just wrong.

PR #340, opened 2019-02-27
Issue 338 - Compare compression against a large sample

Yoric:
This patch introduces examples/sample_directory.rs. This binary takes as argument a directory name
and a sampling factor `f` for the dictionary. From the directory, `f`% of the files are then used
to build a dictionary, while the rest of the files are compressed.

Display the compression results and, optionally, the list of worst offenders.

### DIFF
--- a/examples/sample_directory.rs
+++ b/examples/sample_directory.rs
@@ -1,0 +1,331 @@
+//! Given a directory, generate a dictionary from a random sample, use the
+//! other files in the directory for testing purposes.
+
+extern crate binjs;
+extern crate clap;
+extern crate rand;
+extern crate separator;
+
+use binjs::generic::FromJSON;
+use binjs::io::entropy::dictionary::DictionaryBuilder;
+use binjs::io::entropy::write::Encoder;
+use binjs::io::entropy::Options;
+use binjs::io::{Path as IOPath, Serialization, TokenSerializer};
+use binjs::source::{Shift, SourceParser};
+
+use clap::{App, Arg};
+use rand::{Rng, SeedableRng};
+use separator::Separatable;
+
+use std::fs;
+use std::io;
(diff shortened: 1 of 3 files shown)


examples/sample_directory.rs
```diff
@@ -0,0 +1,324 @@
...
+        results.push((compression, entry.default_path));
+    }
+
+    if show_results {
+        println!("** Ranked results (worse to best)");
+        results.sort_by(|a, b| {
+            f64::partial_cmp(&a.0.binjs_div_brotli, &b.0.binjs_div_brotli).unwrap()
+        });
+        for (compression, path) in results.into_iter() {
+            println!("{:?}\n...original size: {original_size}\n...binjs size: {binjs_size}\n...brotli size: {brotli_size}...ratio: {binjs_div_brotli:.2}\n",
```
RReverser: Could these be using separate `println!` please? There is no difference in output, but would be much more readable in code IMO.

examples/sample_directory.rs
```diff
@@ -0,0 +1,324 @@
...
+            brotli_size,
+            binjs_size,
+            binjs_div_brotli,
+        };
+        results.push((compression, entry.default_path));
+    }
+
+    if show_results {
+        println!("** Ranked results (worse to best)");
+        results.sort_by(|a, b| {
```
RReverser: `.sort_by_key`?

RReverser:
Looks good, left minor nits.

examples/sample_directory.rs
Yoric: Can't `sort_by_key` with `f64`, unfortunately.

RReverser: Ah right. I've used https://github.com/reem/rust-ordered-float to work around that in the past, but maybe extra dependency is not worth it.

PR #344, opened 2019-02-28
Switch benchmark test file to Angular

RReverser:
Now that reusable parser is pretty fast-ish, it's good to benchmark on a real example to take actual Shift conversions into account.


RReverser:
Ah true, that's not the case anymore.

RReverser:
Renamed right here.

Yoric:
Still good for me :)

Issue #342, opened 2019-02-28
Move Shift JSON conversions to JS


RReverser:
Working with untyped JSON is not too pleasant in Rust compared to JS, so I think these conversions are better moved to the JS side.

This would also make it easier to change JSON parser on the Rust side as 1) different parsers have different behaviour when working with untyped JSON values 2) having to use intermediate JSON AST means we can't deserialize typed structures directly from the string.

PR #346, opened 2019-02-28
Move FromShift conversion to JavaScript land

RReverser:
This is the first part of #342 that should make it easier to switch between JSON parsers (or avoid them altogether when used from WASM).

It's mostly manual line-by-line conversion with the exception of using a simpler syntax in JS where possible.

Tests seem to pass, but would love a careful look from someone else as manual conversions tend to be easily error-prone.

Angular parsing benchmark shows ~8% regression in parsing times, but I think we'll easily gain that back longer term by avoiding JSON AST.

### DIFF
--- a/src/source/codegen.js
+++ b/src/source/codegen.js
@@ -8,4 +8,10 @@
 const codegen = require('shift-codegen').default;
 const startJSONStream = require('./start-json-stream');
 
-startJSONStream(ast => codegen(ast));
+startJSONStream({
+    // Mirrors the Rust side of generic value transformations.
+    //
+    // This one takes an AST from the Rust side as an input and applies
+    // a codegen as a transform, returning generated JavaScript back.
+    transform: ast => codegen(ast)
+});
--- a/src/source/from-shift.js
+++ b/src/source/from-shift.js
@@ -1,0 +1,206 @@
+'use strict';
+
+function makeEager(obj) {
+    obj.type = 'Eager' + obj.type;
+}
+
+function dummyDeclaredScope(type) {
+    return {
+        type,
+        declaredNames: [],
+        hasDirectEval: false
+    };
+}
+
+function parameterScopeAndLength(params) {
+    let length = 0;
+    let isSimpleParameterList = true;
+
+    loop: for (let param of params) {
+        switch (param.type) {
+            case 'BindingIdentifier':
+                length++;
+                break;
+            case 'ObjectBinding':
+            case 'ArrayBinding':
+                isSimpleParameterList = false;
+                length++;
+                break;
+            default:
+                isSimpleParameterList = false;
+                break loop;
+        }
+    }
+
+    return {
+        parameterScope: {
+            type: 'AssertedParameterScope',
+            paramNames: [],
+            hasDirectEval: false,
+            isSimpleParameterList
+        },
+        length
+    };
+}
+
+function dummyBoundNamesScope() {
+    return {
+        type: 'AssertedBoundNamesScope',
+        boundNames: [],
+        hasDirectEval: true
+    };
+}
+
+function createFunctionContents(obj) {
+    let kind = obj.type;
+
+    let directives = obj.body.directives || [];
+    delete obj.body.directives;
+
+    let isExpressionBody;
+    let body = obj.body;
+    delete obj.body;
+
+    if (body.type === 'FunctionBody') {
+        body = body.statements;
+        isExpressionBody = false;
+    } else {
+        isExpressionBody = true;
+    }
+
+    let contents;
+
+    switch (kind) {
+        case 'FunctionDeclaration':
+        case 'Method':
+            contents = {
+                type: 'FunctionOrMethodContents',
+                isThisCaptured: false
+            };
+            break;
+        case 'FunctionExpression':
+            contents = {
+                type: 'FunctionExpressionContents',
+                isFunctionNameCaptured: false,
+                isThisCaptured: false
+            };
+            break;
+        case 'ArrowExpression':
+            if (isExpressionBody) {
+                contents = { type: 'ArrowExpressionContentsWithExpression' };
+                obj.type = 'ArrowExpressionWithExpression';
+            } else {
+                contents = { type: 'ArrowExpressionContentsWithFunctionBody' };
+                obj.type = 'ArrowExpressionWithFunctionBody';
+            }
+            break;
+        case 'Getter':
+            contents = {
+                type: 'GetterContents',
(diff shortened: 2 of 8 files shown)


src/source/from-shift.js
```diff
@@ -0,0 +1,205 @@
+'use strict';
+
+function makeEager(obj) {
```
Yoric: Documentation would be really useful, here and in other functions.

src/source/from-shift.js
```diff
@@ -0,0 +1,205 @@
...
+	obj.type = 'Eager' + obj.type;
+}
+
+function dummyDeclaredScope(type) {
```
Yoric: I have the impression that the indentations are much larger than anywhere in the project. Is it me?

I don't really care all that much, but consistency is good :)

src/source/from-shift.js
```diff
@@ -0,0 +1,205 @@
...
+		declaredNames: [],
+		hasDirectEval: false
+	};
+}
+
+function parameterScopeAndLength(params) {
+	let length = 0;
+	let isSimpleParameterList = true;
+
+	loop: for (let param of params) {
```
Yoric: Fun fact: I rewrote the code that handles `break label;` in the SpiderMonkey parser, but I had completely forgotten the existence of this feature :)

src/source/from-shift.js
Yoric: And yes, I realize that the Rust code you're porting didn't have any. Please feel free to blame the authors and/or reviewers of such code, starting with me :)

src/source/from-shift.js
```diff
@@ -0,0 +1,205 @@
...
+	contents.bodyScope = dummyDeclaredScope('AssertedVarScope');
+	contents.body = body;
+
+	obj.contents = contents;
+	obj.directives = directives;
+
+	makeEager(obj);
+}
+
+function convertObject(obj) {
```
Yoric: That's a callback for `JSON.stringify`, right? Can you mention it somewhere?

src/source/from-shift.js
```diff
@@ -0,0 +1,205 @@
...
+	switch (obj.type) {
+		case 'Block':
+			obj.scope = dummyDeclaredScope('AssertedBlockScope');
+			break;
+		case 'BlockStatement':
+			// need to call transform on a nested value manually,
+			// because JSON.stringify will skip it by default
```
Yoric: Er... why?

src/source/codegen.js
```diff
@@ -8,4 +8,6 @@
 const codegen = require('shift-codegen').default;
 const startJSONStream = require('./start-json-stream');
 
-startJSONStream(ast => codegen(ast));
+startJSONStream({
+	transform: ast => codegen(ast)
```
Yoric: `transform` is perhaps not sufficiently clear as a name.

src/source/from-shift.js
RReverser: I blame the authors and/or reviewers, yes :) I'm not sure I'm the best person to describe them as I mostly just transliterated into JavaScript.

src/source/from-shift.js
RReverser: I used tabs here as in other JS files, but yeah, it was me who created other JS files too with different indentation from Rust :/

I don't mind reformatting to spaces if you prefer, but diff will get somewhat larger.

src/source/from-shift.js
RReverser: No, that one is below.

src/source/from-shift.js
RReverser: Well apparently it assumes that we're done with the current value (even if returned object is different), so need to manually re-transform it again in this case.

src/source/codegen.js
RReverser: It mirrors the Rust side of naming. You commented on that one in the Rust PR too, but we didn't come up with something better for a generic "transform arbitrary value to arbitrary value" API.

src/source/from-shift.js
RReverser: Actually, I could've just used fallthrough here...

RReverser: Updated to use fallthrough, I think in this case it's a bit clearer.

src/source/from-shift.js
RReverser: Added comment to the corresponding function.

src/source/from-shift.js
```diff
@@ -0,0 +1,204 @@
...
+			obj.scope = dummyDeclaredScope('AssertedScriptGlobalScope');
+			break;
+		case 'CatchClause':
+			obj.bindingScope = dummyBoundNamesScope();
+			break;
+		case 'StaticPropertyName':
+			obj.type = 'LiteralPropertyName';
+			break;
+		case 'VariableDeclarationStatement':
+			obj = obj.declaration;
```
arai-a: this is doing the same thing as the BlockStatement/Block.
it would be better leaving some comment that `VariableDeclaration` here won't be converted.

src/source/from-shift.js
```diff
@@ -0,0 +1,204 @@
...
+			break;
+	}
+	return obj;
+};
+
+// This is a reviver callback for JSON.stringify to convert Shift AST to BinaryAST JSON.
```
arai-a: why is this called `reviver`?

the parameter name in the spec is `replacer`.
https://tc39.github.io/ecma262/#sec-json.stringify

src/source/codegen.js
Yoric: Can you just comment quickly?

Maybe something as simple as:
```
// Apply `codegen` to the jsonstream.
```

src/source/from-shift.js
Yoric: Fair enough. Can you just file an issue so that we do it eventually?

src/source/codegen.js
RReverser: I'm usually wary of putting comments that don't bring new information not described by the code, but sure. In this case it felt like code is already readable as almost pure English sentence "start JSON stream that transforms AST using codegen".

src/source/from-shift.js
RReverser: You're right, reviver is in JSON.parse. Will change.

src/source/from-shift.js
RReverser: Sure. (just to clarify, I'm not doing it out of principle, I just really don't understand almost half of transformations / helpers here :) )

RReverser:
Fixed nits.

PR #348, opened 2019-03-01
Let investigate_streams rank results by brotli compressed size

Yoric:
Minor change, but it makes prioritization easier.


examples/investigate_streams.rs
```diff
@@ -242,21 +262,25 @@ fn main() {
...
-    for (_, in_extension) in stats {
-        print(in_extension);
+    println!("");
+
+    // Output the rest by `ranking` order.
+    let all_values = Itertools::flatten(stats.into_iter().map(|(_, value)| value.into_iter()));
+    let sorted = match ranking {
+        Ranking::Alpha => all_values.sorted(),
+        Ranking::BrotliCompressedSize => {
+            all_values.sorted_by(|a, b| u64::cmp(&a.1.brotli, &b.1.brotli))
```
RReverser: `sorted_by_key`?

examples/investigate_streams.rs
```diff
@@ -242,21 +262,25 @@ fn main() {
...
+    if let Some(mut in_extension) = stats.remove("binjs") {
+        print_column(&options, "binjs", &in_extension.remove("binjs").unwrap());
     }
-    // Output the rest by alphabetical order.
-    for (_, in_extension) in stats {
-        print(in_extension);
+    println!("");
+
+    // Output the rest by `ranking` order.
+    let all_values = Itertools::flatten(stats.into_iter().map(|(_, value)| value.into_iter()));
```
RReverser: `.flatten()` is part of stdlib these days, so itertools should be unnecessary here.

RReverser: Actually, instead of even `.flatten()` it seems better to replace `.map(...)` with `.flat_map(...)` here.

Yoric: The problem here is that itertools' `flatten` and stdlib's conflict, so I need to desambiguate.

RReverser: What I'm suggesting is specifically to use stdlib's `.flatten()` (or, rather, `flat_map()`).

examples/investigate_streams.rs
```diff
@@ -242,21 +262,23 @@ fn main() {
...
+    if let Some(mut in_extension) = stats.remove("binjs") {
+        print_column(&options, "binjs", &in_extension.remove("binjs").unwrap());
     }
-    // Output the rest by alphabetical order.
-    for (_, in_extension) in stats {
-        print(in_extension);
+    println!("");
+
+    // Output the rest by `ranking` order.
+    let all_values = stats.into_iter().flat_map(|(_, value)| value.into_iter());
```
RReverser: You don't need the inner `.into_iter`, it's implicit due to `flat_map`.

examples/investigate_streams.rs
```diff
@@ -242,21 +262,23 @@ fn main() {
...
+    let sorted = match ranking {
+        Ranking::Alpha => all_values.sorted(),
+        Ranking::BrotliCompressedSize => all_values.sorted_by_key(|x| x.1.brotli),
```
RReverser: Hmm maybe just collect into `Vec` and then do `sort_by_key`? Then you don't need itertools as a dependency here at all.

PR #351, opened 2019-03-07
Unescape escaped WTF-8 properly in binjs_deode (fixes #350)

### LINKED ISSUE #350: WTF-8 escaping is not roundtripping safely
Consider source containing one of the characters that go through WTF-8 escape e.g. `\u007F` generated by:

```js
fs.writeFileSync('temp.js', 'let s = "\u007F;";');
```

Then the `temp.js` will contain:
```js
let s = ";";
```

If you try to encode and decode it again, you will currently get:
```js
let s="007F007F;"
```

I believe this is not the expected outcome and is probably caused by the fact that `escapeWTF8` is used on the ASTs retrieved from Shift parser, but there is no `unescapeWTF8` on the result of Shift codegen.

@arai-a tentatively assigning to you, as I believe you're the most knowledgeable in this area of the codebase.

arai-a:
There were 2 issues:

  - input is not unescaped in `codegen.js`
    - so the output of `codegen` is escaped twice
  - source is not unescaped before writing to file in `decode.rs`
    - so the decoded file is still escaped twice

So:

 - added `inputIsEscapedWTF8` parameter to `start-json-stream.js`
   - if it's `true`, input is unescaped before handling
 - use `inputIsEscapedWTF8` in `codegen.js`
 - unescape the output of `Shift` in `decode.rs`

### DIFF
--- a/src/source/from-shift.js
+++ b/src/source/from-shift.js
@@ -202,7 +202,7 @@ function escapeWTF8(s) {
         if (m == '\u007F') {
             return '\u007F007F';
         }
-        return '\u007F' + m.charCodeAt(0).toString(16);
+        return '\u007F' + m.charCodeAt(0).toString(16).toUpperCase();
     });
 }
 
--- a/tests/test_roundtrip.rs
+++ b/tests/test_roundtrip.rs
@@ -22,9 +22,10 @@ use std::thread;
 /// a `CHANCES_TO_SKIP` probability of being skipped.
 const CHANCES_TO_SKIP: f64 = 0.9;
 
-const PATHS: [&'static str; 2] = [
+const PATHS: [&'static str; 3] = [
     "tests/data/facebook/single/**/*.js",
     "tests/data/frameworks/*.js",
+    "tests/data/misc/*.js",
 ];
 
 fn progress() {
--- a/src/source/shift.rs
+++ b/src/source/shift.rs
@@ -12,6 +12,7 @@ use std::process::*;
 use std::sync::Mutex;
 
 use binjs_generic::syntax::{ASTError, MutASTVisitor, MutASTWalker, WalkPath};
+use binjs_io::escaped_wtf8;
 use binjs_meta::spec::{Interface, NodeName, Spec};
 
 use source::parser::SourceParser;
@@ -158,6 +159,7 @@ impl Shift {
             .transform(&ast)
             .and_then(|mut res| {
                 res.take_string()
+                    .map(escaped_wtf8::to_unicode_escape)
                     .ok_or_else(|| Error::JsonError(json::Error::wrong_type("string")))
             })
             .map_err(|err| {
--- a/tests/data/misc/lone-surrogate.js
+++ b/tests/data/misc/lone-surrogate.js
@@ -1,0 +1,4 @@
+var escape_character = "";
+var lead_surrogate = "<\uD83D>";
+var trail_surrogate = "<\uDC38>";
+var surrogate_pair = "<\uD83D\uDC38>";


src/bin/decode.rs
```diff
@@ -120,10 +120,12 @@ fn main_aux() {
 
     progress!(quiet, "Pretty-printing");
     let printer = Shift::try_new().expect("Could not launch Shift");
-    let source = printer
+    let escaped_source = printer
         .to_source(&spec, json)
         .expect("Could not pretty-print");
 
+    let source = binjs::io::escaped_wtf8::to_unicode_escape(escaped_source);
```
RReverser: This is fixing just the binary, but not the library API. Could it be moved into `to_source`?

src/source/codegen.js
```diff
@@ -13,5 +13,7 @@ startJSONStream({
     //
     // This one takes an AST from the Rust side as an input and applies
     // a codegen as a transform, returning generated JavaScript back.
-    transform: ast => codegen(ast)
+    transform: ast => codegen(ast),
+
+    inputIsEscapedWTF8: true,
```
RReverser: Is this really necessary or would it have the same effect to do `unescapeWTF8(codegen(ast))` in the `transform`?

RReverser: Actually, now I wonder the same for input to parse functions: could we get away with escaping source code itself via WTF-8, and then pass it to the parser, instead of escaping the resulting JSON?

arai-a: I added it there because we already have escapeWTF8 there.

RReverser: Fair enough, but in that case we always unescape, that's why it's in the generic interface. Although, not that I think of it, probably we shouldn't be escaping result of the codegen?

arai-a: can you rephrase?
I don't understand your comment.

RReverser: Yeah sorry, my train of thought just keeps going on...

What I'm thinking about now (past my last comment) is... but can we even realistically unescape WTF-8 and pass it back to Rust?

I mean, Rust expects a valid UTF-8 string from the codegen to write back to the file so it can't really represent surrogates and such anyway and won't be able to serialise these to JS...

RReverser: Oh, I see, you're transforming surrogates back to unicode escapes. I guess that's fine assuming they're inside of string literals.

RReverser: Then back to `escapeWTF8` - I'm just thinking that it's actually wrong to have it in the generic interface as well and needs to be moved to separate module used only by `parse_str` / `parse_file`.

RReverser:
After #352 is merged, could you make a similar change for unescaping in `toJSON` callback of the codegen?

RReverser:
The other PR is merged now. From this one we will likely need just `shift.rs` changes now plus some regression tests added.

src/source/from-shift.js
```diff
@@ -202,7 +202,7 @@ function escapeWTF8(s) {
         if (m == '\u007F') {
             return '\u007F007F';
         }
-        return '\u007F' + m.charCodeAt(0).toString(16);
+        return '\u007F' + m.charCodeAt(0).toString(16).toUpperCase();
```
arai-a: this was necessary to pass roundtrip test which compares AST with escaped WTF-8, because in Rust side we use uppercase for hex string

tests/test_roundtrip.rs
```diff
@@ -22,9 +22,10 @@ use std::thread;
 /// a `CHANCES_TO_SKIP` probability of being skipped.
 const CHANCES_TO_SKIP: f64 = 0.9;
 
-const PATHS: [&'static str; 2] = [
+const PATHS: [&'static str; 3] = [
```
RReverser: I guess it could be easier to use `const PATHS: &[&str] = &[...]` here so that we wouldn't have to update the number anymore, but it's fine as-is.

RReverser:
@arai-a Should I merge it now or do you want to change the thing mentioned above?

PR #352, opened 2019-03-07
Apply escapeWTF8 only on strings in AST

RReverser:
Basically I think it's right to do this first - we shouldn't be applying `escapeWTF8` on result of the codegen in the first place (or on error messages), only on string values in the AST.

This should partially help with #350 but will still need some of your changes in #351.

### DIFF
--- a/src/source/from-shift.js
+++ b/src/source/from-shift.js
@@ -196,10 +196,22 @@ function convertObject(obj) {
     return obj;
 }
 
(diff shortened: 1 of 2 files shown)


RReverser:
Github seems stuck with updating status of `continuous-integration/travis-ci/push` - it has finished several minutes ago, but Github still shows it as pending.

PR #355, opened 2019-03-11
Move BinaryAST -> Shift JSON conversion to JS side 

RReverser:
This is the next step of #342.

Unlike the previous PR (#346), this not only tries to do 1:1 translation from Rust code to JavaScript, but also somewhat simplifies it where possible:

 - Ignores extra properties rather than trying to delete them, since Shift codegen will ignore them as well.
 - Merges all function branches into a catch-all regex.

Also one annoying complication is that JS conversion can't check the interface name it's converting into, so I had to hardcode some parents where `Block` and `VariableDeclaration` are not used as statements. These places are commented and hopefully won't be a big issue to maintain.


src/source/to-shift.js
```diff
@@ -1,31 +1,98 @@
 'use strict';
 
-function removeFunctionContents(obj) {
-    obj.type = obj.type.replace(/^(?:Eager|Lazy)/, '');
+function tryRemoveFunctionContents(obj) {
```
Yoric: Could you document the `try` part? When does it fail? What does it do if it's not a function?

src/source/to-shift.js
```diff
@@ -1,31 +1,98 @@
...
+    let m = obj.type.match(
+        /^(?:Eager|Lazy)(Function(?:Expression|Declaration)|Method|[GS]etter|(ArrowExpression)With(?:FunctionBody|Expression))$/
```
Yoric: I'd be a little more comfortable with a big `switch`. While it's probably a bit slower, it's more readable and more searchable.

src/source/to-shift.js
Yoric: Also, could you add an explanation or example of what this function does?

RReverser: "try" in the sense that it accepts non-functions too (and bails out if it's not a function node).

src/source/to-shift.js
RReverser: Initially I used regex just for trimming the prefix part (it was `/^(?:Eager|Lazy)/`), but later decided that it's easy enough to put types here as well. Can revert.

src/source/to-shift.js
```diff
@@ -1,31 +1,98 @@
...
             directives: obj.directives,
             statements: body
         };
     }
+
+    obj.body = body;
 }
 
-function convertObject(key, value) {}
+function convertObject(key, obj) {
```
arai-a: would be nice to document what `this` is, or receive as `parent` and `child` explicitly.

src/source/to-shift.js
```diff
@@ -1,31 +1,98 @@
...
+            break;
+    }
+
+    return obj;
+}
+
+// This is a reviver callback for JSON.parse to convert Binary AST JSON to Shift AST.
+module.exports = function convert(_key, value) {
+    if (typeof value === 'object' && value !== null && !Array.isArray(value)) {
+        return convertObject.apply(this, arguments);
```
arai-a: `convertObject.call(this, _key, value);` maybe? (to clarify this doesn't expect anything else from `arguments`)

src/source/to-shift.js
```diff
@@ -1,31 +1,98 @@
...
+    } else {
+        return value;
```
arai-a: you don't need `else`

src/source/to-shift.js
RReverser: This one tends to be slightly faster, but it's more of a habit than important difference in this particular case. Will change.

src/source/to-shift.js
RReverser: Yes, but seems clearer that way IMO.

src/source/to-shift.js
arai-a: given this function itself is not passed to `JSON.parse` (also, the call is in different place than this definition), it's not obvious that this is using `JSON.parse` callback interface.
So, it's better explaining that here.
and single line comment won't harm :)

Yoric: Actually, yes, that would be a good place to write that you're inheriting this signature.

src/source/to-shift.js
RReverser: Ah, I recalled why I changed the switch - I didn't really like duplication of each node x2 for both `Eager` and `Lazy` variants, feels hard to read too.

What if I just expand regex to list all types fully instead of using subexpressions?

Or, I guess, I could strip `Eager` / `Lazy` first and then do the switch on the suffix part?

src/source/to-shift.js
```diff
@@ -25,6 +28,10 @@ function tryRemoveFunctionContents(obj) {
     obj.body = body;
 }
 
+// Uses the same signature as JSON.parse callback:
+// - `this` is the parent object being parsed
+// - `key` is the current key in that object
+// - `value` is the value being converted and added to the object.
```
arai-a: `obj`

RReverser: *sigh* right :)

src/source/to-shift.js
RReverser: > Or, I guess, I could strip Eager / Lazy first and then do the switch on the suffix part?

I went with this approach, hopefully it's a good enough balance for readability.

RReverser: *sigh* this broke something, need to recheck.

RReverser: Fixed it.

RReverser:
@Yoric @arai-a Added a commit that also removes a bunch of generic APIs that are not used anymore after this change - can you take a look at that one as well?

crates/binjs_generic/src/syntax.rs
```diff
@@ -1,268 +0,0 @@
...
-use binjs_shared;
-
-use std;
-
-use json::JsonValue as JSON;
-
-pub type WalkPathItem = binjs_shared::ast::PathItem<NodeName, FieldName>;
-pub type WalkPath = binjs_shared::ast::Path<NodeName, FieldName>;
-
-macro_rules! make_ast_visitor {
```
Yoric: Goodbye, `make_ast_visitor`, you will be missed :)

src/source/to-shift.js
```diff
@@ -1,31 +1,130 @@
...
+ */
+function tryRemoveFunctionContents(obj) {
+    let m = obj.type.match(/^(?:Eager|Lazy)(.*)$/);
+    let hasExpressionBody = false;
 
-    delete obj.isAsync;
-    delete obj.scope;
-    delete obj.contents_skip;
-    delete obj.length;
+    switch (m && m[1]) {
```
Yoric: Yeah, I think it's nicer.

PR #356, opened 2019-03-11
 Abstract away AST <-> JSON conversions in Rust

RReverser:
This makes parser and codegen both easier to use and maintain, since the primary public API doesn't rely on a specific JSON parser implementation anymore.

### DIFF
--- a/src/source/shift.rs
+++ b/src/source/shift.rs
@@ -9,7 +9,9 @@ use std::path::*;
 use std::process::*;
 use std::sync::Mutex;
 
+use binjs_es6::ast::Script as ScriptAST;
 use binjs_io::escaped_wtf8;
+use binjs_shared::{FromJSON, FromJSONError, ToJSON};
 
 use source::parser::SourceParser;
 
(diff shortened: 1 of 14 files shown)


RReverser:
Ugh Travis is playing up again - build is finished, but Github status doesn't reflect it.

RReverser:
Force-pushed the same commit, let's see if this helps...

src/source/shift.rs
```diff
@@ -9,7 +9,9 @@ use std::path::*;
 use std::process::*;
 use std::sync::Mutex;
 
+use binjs_es6::ast::Script as ScriptAST;
```
Yoric: Not a big fan of `ScriptAST` as a name. Maybe `ES6ScriptRoot`?

RReverser: Maybe... Could as well just use `AST` or just import `binjs_es6::ast` and use it as `ast::Script` everywhere. Idk.

RReverser: Or maybe I should rename local `Script` to something like `Bindings`?

Yoric: Yeah, let's call it `AST` for the moment. If we ever extend this to other languages (e.g. JSON), we'll have time to change name.

RReverser: Done.

PR #358, opened 2019-03-12
Run only subset of testcases in roundtrip (fixes #200)

### LINKED ISSUE #200: Remove randomness from test_roundtrip
having randomness in testsuite is bad :P
it tends to overlook failure.

arai-a:
- Chose only 4 files from dataset (2 from facebook, 2 from framework).
 - Removed randomness
 - Test files are still there


arai-a:
okay, now whole build+test takes only 17minutes

PR #361, opened 2019-03-14
Raw `json` -> `serde_json` replacement

RReverser:
This is first part of #357 that intentionally tries to perform only 1:1 raw replacement where possible, with minor changes for incompatible APIs.

Check commit history for details - first few commits adjust usages to universal APIs to support either `json` or `serde_json` alike, and the last one does the actual replacement of one crate with another.

I'll send a PR that removes intermediate JSON AST representation after this one is merged, to make reviews easier and more scoped.


crates/binjs_shared/src/json_conversion.rs
```diff
@@ -26,30 +28,18 @@ impl FromJSON for bool {
...
-                }
-            }
-            _ => Err(FromJSONError {
-                expected: "Number".to_string(),
-                got: value.to_string(),
-            }),
+        if value.is_null() {
+            return Ok(std::f64::NAN);
         }
+        value.as_f64().ok_or_else(|| FromJSONError {
```
arai-a: nice that the hack for manual conversion gets removed :)
can you remove `JSON.as_f64() doesn't keep the precision.` comment above?

RReverser:
Oh, sure.

crates/binjs_io/src/binjs_json/write.rs
```diff
@@ -112,24 +111,30 @@ impl TokenWriter for TreeTokenWriter {
                 interface: tag,
                 field_values,
             }) => {
-                let mut obj = json::object::Object::with_capacity(2);
-                obj.insert("@TYPE", JSON::String("tagged tuple".to_string()));
-                obj.insert("@INTERFACE", JSON::String(tag.as_str().to_string()));
+                let mut obj = serde_json::Map::with_capacity(2);
```
Yoric: Couldn't this be

```rust
let mut obj = json!{
  "@TYPE": "tagged tuple",
  "@INTERFACE": tag
};
```
?

Also in a few places below.

tests/test_annotations.rs
```diff
@@ -27,47 +27,47 @@ test!(test_annotations_scopes_1, {
         .expect("Could not annotate AST");
 
     // Toplevel
-    assert_eq!(ast["BINJS:Scope"], object!{
+    assert_eq!(ast["BINJS:Scope"], json!({
         "type" => "BINJS:Scope",
```
Yoric: Does serde understand `=>` in `json!`?

crates/binjs_io/src/binjs_json/write.rs
RReverser: It could! But as I mentioned in the description, I intentionally tried to make this 1:1 replacement without refactorings for easier reviews and validation :)

tests/test_annotations.rs
RReverser: Hmm good catch. I don't think it does, but not sure how these tests pass either...

crates/binjs_io/src/binjs_json/write.rs
Yoric: Sounds good :)

tests/test_annotations.rs
RReverser: Okay, it's definitely not supported. I'm confused. Does this test even run?

RReverser: Oh right, it's all commented out...

RReverser: @Yoric Fixed but TBH it feels weird to maintain large commented out block of code... Can we either uncomment it out (if it works) or delete and rely on Git history should we ever want to restore it?

crates/binjs_shared/src/json_conversion.rs
RReverser: Thanks, done.

tests/test_annotations.rs
Yoric: You're right, that is weird.

Could you quickly check whether it works? Otherwise, let's get rid of it.

RReverser: It doesn't, mostly broken by the fact that `parse_str` now returns statically typed objects and not JSON.

RReverser: But even before that, I think the JSON keys used weren't valid for a while now.

RReverser: Removed test in a separate commit and rebased branch on top of it.

RReverser:
Huh, some new failures after rebase.

PR #364, opened 2019-03-19
Introducing trait WalkGuard

Yoric:
This should simplify a bit the code.

Part on the ongoing work towards specializing for JSON.

### DIFF
--- a/crates/binjs_generate_library/src/lib.rs
+++ b/crates/binjs_generate_library/src/lib.rs
@@ -333,7 +333,7 @@ impl<W> Serialization<{name}> for Serializer<W> where W: TokenWriter {{
                 let walker = format!("
 impl<'a> Walker<'a> for {name} where Self: 'a {{
     type Output = {name};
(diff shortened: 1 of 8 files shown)


crates/binjs_generate_library/src/lib.rs
```diff
@@ -1511,8 +1516,18 @@ pub trait Visitor<E, G=()> where G: for<'me> From<&'me Self> {{
 }}\n
 pub trait Walker<'a>: Sized {{
     type Output;
-    fn walk<V, E, G>(&'a mut self, path: &mut WalkPath, visitor: &mut V) -> Result<Option<Self::Output>, E> where V: Visitor<E, G>, G: for<'w> From<&'w V>;
+    fn walk<V, E, G>(&'a mut self, path: &mut WalkPath, visitor: &mut V) -> Result<Option<Self::Output>, E> where V: Visitor<E, G>, G: WalkGuard<V>;
 }}\n
+pub trait WalkGuard<V> where Self: Sized {{
+    fn new(visitor: &V, path: &WalkPath) -> Self;
+}}\n
```
arai-a: just noticed
why is this using escape sequence instead of raw newline?

Yoric: Oops, fixed.

PR #365, opened 2019-03-19
Introducing a new enum variant BinASTStolen to represent stolen AST nodes

Yoric:
The ViewMut* mechanism lets clients *steal* subnodes of a tree for rewriting purposes.
To steal a subnode, it needs to create a `Default::default()` node, but this can lead
to infinite recursions if the default case happens to rely upon a node of the same type.

This patch introduces `Foo::BinASTStolen` for all AST enums `Foo`, which serves to
generate the default node, and serves whenever we steal a node.

### DIFF
--- a/crates/binjs_generate_library/src/lib.rs
+++ b/crates/binjs_generate_library/src/lib.rs
@@ -332,7 +332,7 @@ impl<W> Serialization<{name}> for Serializer<W> where W: TokenWriter {{
                 let walker = format!("
 impl<'a> Walker<'a> for {name} where Self: 'a {{
     type Output = {name};
-    fn walk<V, E, G>(&'a mut self, _: &mut WalkPath, _: &mut V) -> Result<Option<Self::Output>, E> where V: Visitor<E, G>, G: for<'w> From<&'w V> {{
+    fn walk<V, E, G>(&'a mut self, _: &mut WalkPath, _: &mut V) -> Result<Option<Self::Output>, E> where V: Visitor<E, G>, G: WalkGuard<V> {{
         Ok(None)
     }}
 }}\n",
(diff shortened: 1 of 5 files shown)


crates/binjs_generate_library/src/lib.rs
```diff
@@ -321,12 +321,12 @@ impl<W> Serialization<{name}> for Serializer<W> where W: TokenWriter {{
...
                         .iter()
                         .map(|s| format!(
-                            "            {name}::{typed} => \"{string}\"",
+                            "            {name}::{typed} => \"{string}\",",
                             name = name,
                             typed = s.to_cpp_enum_case(),
                             string = s,
                         ))
-                        .format(",\n")
+                        .format("")
```
arai-a: isn't it nice to keep newline?

crates/binjs_generate_library/src/lib.rs
```diff
@@ -442,10 +446,10 @@ pub enum ViewMut{name}<'a> {{\n{ref_mut_contents}\n}}\n",
                         contents = types
                             .iter()
                             .map(|case| format!(
-                                "    {name}(Box<{name}>)",
+                                "    {name}(Box<{name}>),",
                                 name = case.to_class_cases()
                             ))
-                            .format(",\n"),
+                            .format(""),
```
arai-a: here too.

others have newline in each item, but above two don't.

crates/binjs_generate_library/src/lib.rs
```diff
@@ -515,10 +520,7 @@ impl Default for {name} {{
 
 ",
                         name = name,
-                        default = format!(
-                            "{variant}(Box::new(Default::default()))",
-                            variant = types[0].to_class_cases()
-                        ),
+                        default = "BinASTStolen",
```
arai-a: this doesn't have to be parameter, but just embed into the above code?

PR #367, opened 2019-03-20
Migrate from FromJSON / ToJSON to Serde

### LINKED ISSUE #357: Migrate to Serde for generic deserialisation from JSON and more

RReverser:
This is migrating from custom `FromJSON` / `ToJSON` to generic Serde implementations, thus streamlining parsing from input strings directly into typed Rust structures with no intermediate JSON AST.

Fixes #357. 

This is mostly handled by `#[derive(Serialize, Deserialize)]` with the exception of tagged enums in AST for which we now generate manual `Deserialize` implementation to avoid a slow path with buffering in `serde-derive` (see comments and/or commit message for the explanation).

On my machine and the current Angular parsing benchmark this results in:
 - CPU improvements: ~157ms -> ~105ms (-33% or 1.5x)
 - RAM improvements: 44.7 MB -> 7.0 MB (-84% or 6.4x) - this one looks almost too good to be true, but rechecked several times and both numbers are pretty consistent.


crates/binjs_generate_library/src/lib.rs
```diff
@@ -429,13 +386,49 @@ impl<'a> Walker<'a> for {name} where Self: 'a {{
...
-#[derive(PartialEq, Debug, Clone)]
+#[derive(PartialEq, Debug, Clone, Serialize)]
+#[serde(tag = \"type\")]
 pub enum {name} {{
 {contents}
     /// An additional value used to mark that the node was stolen by a call to `steal()`.
     BinASTStolen,
 }}\n
 
+// An optimised implementation of tagged deserialise that expects `type` to be the first key in the object.
```
Yoric: Won't that make the deserializer incompatible with many tools and possibly with future versions of Shift?

How difficult would it be to implement a fallback that goes the slow way? Not necessarily in this patch, but maybe in a followup.

RReverser: Yeah, in theory it could, but:

1. We currently have only one well-defined source from which we currently deserialise these ASTs.
2. In practice I haven't seen real-world cases where tagged enums wouldn't have this `type` field as first in the object, especially if it comes from well-defined typed structures (like Shift AST or Rust `derive(Serialize)`).
3. Implementation raises a clear error if this contract is violated.

If someday we deserialise AST from a source where `type` field does happen not to be the first one in the JSON, we'll notice a clear error from (3) right away and might fix by either going back to `derive(Deserialize)` or modifying the transform on the JS side or buffering into an intermediate AST ourselves.

Until then, given how bad this slow path is, it seems better to raise an error than try to silently accept it.

Btw, later turned out that comment might be not 100% correct - I've been pointed out that JSON spec does allow implementors to assign semantic meaning to the order of keys, so we don't violate it. https://twitter.com/awbjs/status/1106287165892288512

Yoric: Works for me.

crates/binjs_generate_library/src/lib.rs
```diff
@@ -429,13 +386,49 @@ impl<'a> Walker<'a> for {name} where Self: 'a {{
...
+        enum VariantTag {{
+            {variant_tags}
+        }}
+
+        struct MapVisitor;
+
+        impl<'de> serde::de::Visitor<'de> for MapVisitor {{
+            type Value = {name};
+
+            fn expecting(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {{
```
Yoric: I don't see where this function is used.

Yoric: Also, what exactly *is* the error message if the type doesn't come first?

RReverser: It's implementation of Serde trait. It's not used directly by us.

RReverser: Just checked and tried to pass invalid JSON:

> `Could not parse source: JSONError(Error("unknown variant `expression`, expected `type`", line: 1, column: 3205177))`

I guess it could be better, I just was lazy and used a simple `enum` with a single variant.

What do you think - should I implement custom deserialize there to provide a better message?

Yoric: If it's not complicated, yeah, that would be nicer.

Because the day we notice this error message, I'm sure that I will have forgotten what it means :)

RReverser: Yeah, agreed.

RReverser: Updated to:

> Could not parse source: JSONError(Error("invalid value: key `expression`, expected a `type` key as the first item in the object", line: 1, column: 3205177))

RReverser:
I just realised that top-level tagged `Result` still does its own buffering of entire JSON into an intermediate AST, so there are further optimisations to unlock.

crates/binjs_generate_library/src/lib.rs
Yoric: Much better, thanks!

Perhaps emphasize word `first`?

RReverser: Like how?

Yoric: Like `*first*` or `FIRST` :)

RReverser:
> I just realised that top-level tagged `Result` still does its own buffering of entire JSON into an intermediate AST, so there are further optimisations to unlock.

Actually, just checked and this one doesn't make any difference - it looks like an adjacently-tagged (`tag` + `content`) shape which I'm using for `Result` already has a fast path for `tag` being the first key.

I hope to upstream my optimisations for internally-tagged enum in a generic way to Serde, but for now I'll leave this as-is.

PR #379, opened 2019-03-25
Making annotations more robust

Yoric:
We don't necessarily want to panic whenever we cannot enrich a file. This patch is a first step towards converting our Enrich mechanism to something that bails out gracefully whenever it encounters an error.

### DIFF
--- a/crates/binjs_es6/src/lib.rs
+++ b/crates/binjs_es6/src/lib.rs
@@ -26,8 +26,12 @@ pub mod io;
 /// which we do not support yet.
 #[derive(Debug)]
 pub enum EnrichError {
-    /// While analyzing scopes, we exit a binding identifier but there is no binding kind.
-    MissingBindingKind,
+    ScopeError(scopes::ScopeError),
+}
+impl From<scopes::ScopeError> for EnrichError {
+    fn from(err: scopes::ScopeError) -> Self {
+        EnrichError::ScopeError(err)
+    }
 }
 
 /// A mechanism used to enrich an AST obtained from parsing with additional information
--- a/crates/binjs_es6/src/lib.rs
+++ b/crates/binjs_es6/src/lib.rs
@@ -4,10 +4,12 @@
 
 #[macro_use]
 extern crate binjs_io;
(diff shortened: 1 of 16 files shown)


crates/binjs_es6/src/lib.rs
```diff
@@ -55,17 +57,70 @@ pub struct Enrich {
...
+            Arg::with_name("inject-lazy")
+                .long("inject-lazy")
+                .takes_value(true)
+                .value_name("LEVEL")
+                .help("Rewrite the AST for all functions strictly below level LEVEL. Do nothing if LEVEL = 0")
+                .default_value(defaults_as_strings::DEFAULT_LAZY_THRESHOLD),
+            Arg::with_name("inject-scopes")
+                .long("inject-scopes")
+                .takes_value(true)
+                .help("Rewrite the AST to provide scope information.")
```
arai-a: would be nice to say "(defaults to true)"
(or is it added automatically?)

crates/binjs_es6/src/lib.rs
```diff
@@ -55,17 +57,70 @@ pub struct Enrich {
...
 }
 impl Enrich {
+    pub fn args<'a, 'b>(&self) -> Vec<clap::Arg<'a, 'b>> {
+        use clap::*;
+        vec![
+            Arg::with_name("inject-lazy")
+                .long("inject-lazy")
+                .takes_value(true)
+                .value_name("LEVEL")
+                .help("Rewrite the AST for all functions strictly below level LEVEL. Do nothing if LEVEL = 0")
```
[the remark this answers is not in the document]
Yoric: Yes, it's the exact same thing as `--lazify`. Adding an alias so that we can write either `--lazify` or `--inject-lazy`.

crates/binjs_es6/src/lib.rs
Yoric: It is added automatically.

Yoric:
I have added an alias that will let us use `--lazify` for `--inject-lazy`.

PR #382, opened 2019-03-28
Doc: Including entropy.md from entropy/mod.rs

Yoric:


crates/binjs_io/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![feature(external_doc)]
```
RReverser: This will make it impossible to build crate on stable AFAIK.

RReverser: One thing you can do is hide it under a feature flag which will be passed only when building docs using nightly Cargo, e.g.:

```
#![cfg_attr(feature = "unstable", feature(external_doc))]
```

(with corresponding feature declaration in `Cargo.toml` and passing it in a Travis script)

Yoric: Normally, this should now be fixed.

PR #388, opened 2019-04-03
Create template for context-0.2 documentation as a copy of entropy.md

arai-a:
Copied entropy.md to context.md (temporary name), and modified some strings to make the implementation simpler.
the format itself is still for entropy, and contains unnecessary parts for the new format.
those parts will be removed/changed once we get the code.

### DIFF
--- a/spec/context.md
+++ b/spec/context.md
@@ -2,6 +2,11 @@
 
 This is an attempt to document the context-0.2 compression format.
 
+WARNING: the content is almost just a copy of entropy.md and doesn't reflect
+the actual format.  They're about to be changed so much once we fixed the
+format.  The current content is here to clarify the difference between entropy
+in the future changes.
+
 # Global structure
 
 ```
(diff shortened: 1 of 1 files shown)


spec/context.md
```diff
@@ -0,0 +1,392 @@
...
+
+```
+Prelude ::= "PREL" NamedPreludeDictionary*
+NamedPreludeDictionary ::= "FLOA" PreludeStream<f64?>
+                         | "ULON" PreludeStream<u32?>
+                         | "LIST" PreludeStream<u32?>
+                         | "PROP" PreludeStream<String>
+                         | "IDEN" PreludeStream<String>
+                         | "STRI" PreludeStream<String>
+                         | "PROp" PreludeStream<StringLen>
```
Yoric: `PROP` vs. `PROp` is not very readable.

Maybe `PROP` vs. `prop`? `PROC` vs. `PROL` (`C` = content, `L` = len)?

arai-a: thanks. I'll use C/L

PR #389, opened 2019-04-03
Context 0.2: Prelude decoder

Yoric:
This patch adds a PreludeDecoder for Context 0.2 and exposes to Context 0.2
the other decoders developed for Entropy 0.4 and that may be reused as such in Context 0.2.

### DIFF
--- a/crates/binjs_io/src/context/prelude.rs
+++ b/crates/binjs_io/src/context/prelude.rs
@@ -1,0 +1,215 @@
+use context::Name;
+use TokenReaderError;
+
+use bytes::varnum::ReadVarNum;
+use util::PosRead;
+
+use std::io::Read;
+
+/// A compression format used in the prelude.
+#[derive(Clone, Debug)]
+pub enum Compression {
+    /// The subsection is compressed using brotli.
+    Brotli,
+    /// The subsection is uncompressed.
+    Identity,
+}
+impl Compression {
+    /// Return the representation as a 4 bytes name.
+    pub fn as_bytes(&self) -> &'static Name {
+        match *self {
+            Compression::Brotli => b"BROT",
+            Compression::Identity => b"NONE",
+        }
+    }
+
+    /// Parse the 4 bytes name.
+    pub fn from_bytes(bytes: &Name) -> Option<Compression> {
+        match bytes {
+            b"BROT" => Some(Compression::Brotli),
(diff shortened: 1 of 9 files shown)


Yoric:
Sorry, I had forgotten to `git add` roughly half of the code :/

Could you re-review this PR?

crates/binjs_io/src/context/prelude.rs
```diff
@@ -0,0 +1,214 @@
...
+                return Ok(None);
+            }
+            Some(section) => section,
+        };
+
+        // Read compression format.
+        let mut name = [0; 4];
+        self.reader
+            .read_exact(&mut name)
+            .map_err(TokenReaderError::ReadError)?;
```
arai-a: not about this code but the format itself,
is there any reason that we have compression format specifier in each stream?
there can be more than 10 streams in single file, and I don't see any reason to use different compression for each stream.
(at least for streams inside prelude/content)
are you planning to change it in the future, to use single compression type for prelude/content/main for each, or something along that line?

crates/binjs_io/src/context/prelude.rs
```diff
@@ -0,0 +1,214 @@
...
+            b"BROT" => Some(Compression::Brotli),
+            b"NONE" => Some(Compression::Identity),
+            _ => None,
+        }
+    }
+}
+
+/// The name of a section of the prelude.
+#[derive(Clone, Debug)]
+pub enum Section {
```
arai-a: I was thinking that prelude/content/main are called Section, and each dictionary is something like SubSection.
is it wrong?

Yoric: Oops, SubSection, you're right.

crates/binjs_io/src/context/prelude.rs
Yoric: The original reason was that I have witnessed cases in which Brotli made some streams quite larger for small files. I'm not 100% sure it's a good idea, though.

Yoric: Maybe we'll get rid of this in a future version?

arai-a: oh, indeed.
encoder can choose uncompressed format when brotli "compressed" stream becomes larger than original.
let's keep it as is for now :)

PR #391, opened 2019-04-04
Let's try to specify chunks

Yoric:


spec/context.md
```diff
@@ -56,10 +56,39 @@ The shared dictionary is optional.
...
+ChunkNum ::= var_u32
+LazyNum ::= var_u32
+```
+
+A `ChunkIdentifier` may be:
+
+- `EAGR`, in which case the chunk contains toplevel code, meant to be parsed and executed immediately;
+- `LAZY`, in which case the chunk contains lazy function definitions, meant to be parsed/executed by-need.
+
+It is a syntax error for an `EAGR` chunk to appear after a `LAZY` chunk.
```
arai-a: So, when concatenating multiple files both with eager and lazy chunks, they should be sorted, right?

Yoric: Yes. I figure that it should be fairly easy to do. If the `ChunkNum` do not collide, it's just taking 4 ranges of bytes and reordering them.

Note: Maybe `ChunkNum` should be a fixed length number, rather than a `var_u32`.

Yoric:
Merging without waiting for tests: it's just a spec document.

PR #392, opened 2019-04-04
Fix order of flags for cargo doc

RReverser:


Yoric:
#393 is more complete. Sorry for the race to patch :)

PR #393, opened 2019-04-04
Doc: Fixing feature 'unstable', used to export doc spec

Yoric:


crates/binjs_io/src/entropy/format.rs
```diff
@@ -1,3 +1,6 @@
 // Fake module. This is only a placeholder for documentation.
 
+// As of this writing, doc inclusion only works on Nightly compiler.
 #![cfg_attr(feature = "unstable", doc(include = "../../../spec/entropy.md"))]
+
+#![cfg_attr(not(feature = "unstable"), doc = "To generate the format spec documentation, please use `cargo doc --features unstable`. This requires Rust Nightly.")]
```
RReverser: I'm not sure what is this doing?

RReverser: Ah... is this just adding a comment to the generated docs when on stable?

Yoric: It should generate a placeholder documentation when people build the doc with a simple `cargo doc`. Not strictly necessary, but at least people won't end up with an empty module without any documentation :)

PR #395, opened 2019-04-04
Context 0.2: Prelude Encoder

Yoric:


crates/binjs_io/src/context/prelude.rs
```diff
@@ -31,10 +46,19 @@ impl Compression {
             _ => None,
         }
     }
+
+    /// Return a file extension for this compression format.
+    /// Used typically for dumping samples.
+    pub fn extension(&self) -> &'static str {
+        match *self {
+            Compression::Brotli => "bro",
```
arai-a: "br" maybe?
https://github.com/google/ngx_brotli#brotli_static

Yoric: Works for me.

I seem to remember used to produce `.bro` files, then the word "bro" became associated to sexual harassment (I think) in the US and they renamed it to `.br`.

Anyway, it's a simple enough patch :)

PR #402, opened 2019-04-14
Allow the Node.js encoder source path to be specified via environment variable

vdjeric:
Currently the location of the Node.js encoder files is hard-coded at build-time. This patch allows it to be specified via environment variable.

I'll squash before committing this pull request.


Yoric:
I believe I understand the point of the patch, but I think that we should find a better solution. Right now, we can't deploy the binary without the source file. Perhaps the best thing to do would be to somehow embed the source files in the binary?

@RReverser, any idea? How do you deploy at CloudFlare?

RReverser:
I was thinking about this problem too, but I don't have a good answer yet. This PR seems reasonable, as well as suggestion to embed source files into the binary, although that would require unpacking them to a temp dir before starting Node.js (in `Shift::try_new`).

vdjeric:
I get the appeal of embedding JS source files in the binary, but I also think it's too much static-linking. We may want to change the transform code without modifying the Rust binary. 
Let's land this patch for now and look for a longer-term solution?

RReverser:
>  I also don't like the idea of creating new temp files for encoding every single Facebook JS module (we have 10s of thousands)

This wasn't suggested, result of `Shift::try_new` can and should be reused for many invocations to speed up compilation without restarting Node.js process over and over. Auto-deleted temporary files wouldn't add much overhead in that case, as they would be created only once at the start of the process.

But yeah, landing as-is for now sounds good.

Yoric:
Let's do that.

PR #404, opened 2019-04-16
Spec: First stab at specifying Grammar

Yoric:


spec/context.md
```diff
@@ -13,7 +13,66 @@ This document specifies how to compress an Abstract Syntax Tree (or AST) in a gi
...
+
+```typescript
+type Grammar = Set<Declaration>;
+
+class Declaration {
+  name: String;
+  content: Content;
+};
+class Content {
+  optional: bool;
```
arai-a: what is `optional` field for?

Yoric: For optional values. I should document this.

spec/context.md
```diff
@@ -13,7 +13,66 @@ This document specifies how to compress an Abstract Syntax Tree (or AST) in a gi
...
+           | Interface;
+
+class PrimitiveString { }
+class PrimitiveF64 { }
+class PrimitiveU32 { }
+class PrimitiveBool { }
+class PrimitivePropertyKey { }
+class PrimitiveIdentifierName { }
+class Array {
+  values: Definition
```
arai-a: shouldn't this be `Content` (if nullable) or `Value` (non-nullable) ?
I don't think it elements should be named.

Yoric:
Closing PR. We might resurrect this at some point after #428 has landed.

PR #407, opened 2019-05-02
Resolve #406 - exit_binding_identifier now handles lazy functions

### LINKED ISSUE #406: Lazification broken on master
Looks like binast/binjs-ref@1b17153 this commit broke lazification.

Literally the one before it works, but trying to lazify anything with this one results in panics with `MissingBindingKind` error.

I guess this wasn't caught because we don't lazify things in tests.

Yoric:
exit_binding_identifier contains a hard-coded list of nodes that represent
functions. This is hardly ideal (perhaps we should somehow put this in the webidl?)
and at some point, we failed to update this list, causing the bug.

### DIFF
--- a/crates/binjs_es6/src/scopes.rs
+++ b/crates/binjs_es6/src/scopes.rs
@@ -519,6 +519,75 @@ impl AnnotationVisitor {
     fn pop_function_expression_name(&mut self) {
         self.function_expression_name_stack.pop();
     }
+
+    fn exit_lazy_or_eager_function_declaration<T>(
+        &mut self,
+        path: &WalkPath,
+        name: &BindingIdentifier,
+    ) -> ExitResult<T> {
+        // If a name declaration was specified, remove it from `unknown`.
+        let name = name.name.clone();
+
+        // Var scope is already committed in exit_function_or_method_contents.
+        // The function's name is not actually bound in the function; the outer var binding is used.
+        // Anything we do from this point affects the scope outside the function.
+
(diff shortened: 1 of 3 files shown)


crates/binjs_es6/src/scopes.rs
```diff
@@ -1153,81 +1174,39 @@ impl Visitor<ScopeError> for AnnotationVisitor {
...
+        Ok(None)
+    }
+
     fn exit_eager_function_declaration(
         &mut self,
         path: &WalkPath,
         node: &mut EagerFunctionDeclaration,
     ) -> ExitResult<EagerFunctionDeclaration> {
         debug!(target: "annotating", "exit_eager_function_declaration {:?} at {:?}", node.name.name, path);
+        self.exit_lazy_or_eager_function_declaration(path, &node.name)?;
```
RReverser: In all of these, it seems you could just write:

```rust
self.exit_lazy_or_eager_function_declaration(path, &node.name)
```

and return result immediately?

Yoric: The return type is different, unfortunately.

RReverser: Ah. I guess you can make that function generic to return any `ExitResult<T>`?

RReverser:
Seems fine by me, but I'll better leave it to you and @arai-a :)

RReverser:
@Yoric Although, could you maybe change or re-enable tests responsible for lazification as part of this?

PR #408, opened 2019-05-02
Replace set_extension calls

xtuc:
create-react-app generate files with [name].[hash].js format. Rust's
Path considers that [hash].js is the extension, set_extension loses the
hash.

### DIFF
--- a/src/bin/encode.rs
+++ b/src/bin/encode.rs
@@ -247,10 +247,7 @@ fn main_aux() {
                 .short("q")
                 .help("Do not print progress"),
         ])
-        .args(Enrich {
-            scopes: true,
-            ..Default::default()
-        }.args().as_slice())
+        .args(Enrich::default().args().as_slice())
         .subcommand(binjs::io::Format::subcommand())
         .get_matches();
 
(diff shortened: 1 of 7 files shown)


src/bin/encode.rs
```diff
@@ -92,12 +92,10 @@ fn handle_path<'a>(options: &mut Options<'a>, source_path: &Path, sub_dir: &Path
                 .expect("Could not find or create destination directory");
 
             let mut bin_path = d.join(sub_dir);
-            bin_path.push(file_name);
-            bin_path.set_extension("binjs");
+            bin_path.push(file_name.to_string_lossy().to_string() + ".binjs");
```
RReverser: It might be better to do `format!("{}.binjs", file_name.to_str().unwrap())` - aside from efficiency differences, this will allow to fail on cases when filename contains unsupported characters rather than silently replace them.

In the future we might want to support them properly, but it's unlikely to be a common case, so just failing would be good for now.

xtuc:
I suspect that `.{grammar,strings,tree}` files are still using `set_extension`.

RReverser:
@xtuc Possible. I guess let's fix all of them together.

xtuc:
I just tested it, it's working as expected.

xtuc:
@RReverser could you please another look at this?

PR #413, opened 2019-05-25
Update itertools.

emilio:
This allows me to update it in Gecko without duplicating the dependency.


crates/binjs_generate_library/src/lib.rs
```diff
@@ -347,9 +347,8 @@ impl<'a> Walker<'a> for {name} where Self: 'a {{
             let mut options = vec![];
             let mut lists = vec![];
             let mut primitives = vec![];
-            let names: Vec<_> = source.keys().sorted();
 
-            for name in names.into_iter() {
+            for name in source.keys().sorted() {
```
Yoric: Oh, they changed the return type of `sorted()`? Fun times.

emilio:
Yeah, a new version would be appreciated! :)

Yoric:
@emilio it doesn't seem to pass tests, can you check?

emilio:
I had missed a `--all` in my cargo test command. Should be good now :)

Yoric:
Thanks :)

Yoric:
New version released.

Yoric:
@emilio I'm updating SpiderMonkey to use the latest version but `./mach rust vendor` still claims that 0.7.x is used somewhere.

emilio:
Yes, it's used in the style system, I can take care to update it and you can review, or vice versa I guess :)

emilio:
As far as I can tell, updating itertools in the style crate should require no code changes at all, so I'm happy to rubber-stamp the version increment.

PR #422, opened 2019-08-08
bpy: Added a new subcommand encode-dir to encode entire directories

Yoric:
This will simplify generating large batches of tests.

### DIFF
--- a/fbssdc/bpy.py
+++ b/fbssdc/bpy.py
@@ -8,13 +8,47 @@
 import format
 import idl
 import opt
+import os
 import strings
+import subprocess
+import tempfile
 import tycheck
 import types
 
 import argparse
 import json
 import sys
+
+def encode_dir(dict_file, binjs_encode, in_path, out_path):
+  types = idl.parse_es6_idl()
+  ty_script = types.interfaces['Script']
+  string_dict = strings.read_dict(dict_file, with_signature=True)
+  in_path = os.path.abspath(in_path)
+  out_path = os.path.abspath(out_path)
+  ignored_out_directory = tempfile.TemporaryDirectory()
+  for root, _, sources in os.walk(in_path):
+    # 1. Prepare destination directory
+    suffix = os.path.relpath(root, in_path)
+    dest_root = os.path.join(out_path, suffix)
+    print ('Encoding from {root} to {dest_root}'.format(root=root, dest_root=dest_root))
+    os.makedirs(dest_root, exist_ok=True)
+
+    for source in sources:
+      source_path = os.path.join(root, source)
+      if not source[-3:] == '.js':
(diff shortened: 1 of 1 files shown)


fbssdc/bpy.py
```diff
@@ -7,15 +7,51 @@
...
+  types = idl.parse_es6_idl()
+  ty_script = types.interfaces['Script']
+  string_dict = strings.read_dict(dict_file, with_signature=True)
+  in_path = os.path.abspath(in_path)
+  out_path = os.path.abspath(out_path)
+  ignored_out_directory = tempfile.TemporaryDirectory()
+  for root, _, sources in os.walk(in_path):
+    # 1. Prepare destination directory
+    suffix = os.path.relpath(root, in_path)
+    print ("Root: {root}, suffix: {suffix}, out_path: {out_path}".format(root=root,suffix=suffix,out_path=out_path))
```
vdjeric: isn't this more of a debug statement?

fbssdc/bpy.py
```diff
@@ -7,15 +7,51 @@
 
 import format
 import idl
+import io
```
vdjeric: do you use this module anywhere?

fbssdc/bpy.py
```diff
@@ -7,15 +7,51 @@
...
+
+    for source in sources:
+      source_path=os.path.join(root, source)
+      if not source[-3:] == ".js":
+        print ("...skipping {}".format(source_path))
+        continue
+
+      # 2. Extract AST
+      print ("Preprocessing {}".format(source_path))
+      process = subprocess.run([binjs_encode, "--quiet", "--show-ast", "--in", source_path, "--out", ignored_out_directory.name], capture_output=True)
```
vdjeric: nit: you're a bit inconsistent about using single vs double quoted strings

fbssdc/bpy.py
```diff
@@ -89,6 +125,23 @@ def main():
...
+
+  encode_full_parser = subs.add_parser('encode-dir', help='Encode a full directory (from .js).',
+                                  description='''Caveats:
+(1) dictionary misses are not supported yet
+(2) not all file sections are compressed natively yet;
+    the output should be compressed with Brotli
+''')
+  encode_full_parser.add_argument('binjs_encode',
+                             help='path to the binjs_encode binary')
+  encode_full_parser.add_argument('dictionary', type=argparse.FileType('rb'),
```
vdjeric: what do you think of making dictionary arg optional so that it can default to an empty external dictionary

Yoric: Maybe in a followup bug. For the moment, I don't think that `format.write` supports dictionary as an optional argument.

PR #430, opened 2019-09-02
Context 0.1: Implementation of varnum reading

Yoric:
Ok, let's get porting context 0.1 to Rust started!


crates/binjs_io/src/context/varnum.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+#[test]
+fn test_read_varu32_no_normalization_must_work() {
+    use std::io::Cursor;
+    for (input, expected) in &[
+        // Various versions of 0
+        (vec![0], 0),
+        (vec![0b10000000, 0b00000000], 0),
+        (vec![0b10000000, 0b10000000, 0b00000000], 0),
```
dominiccooney: Maybe also test a couple of multi-length, denormalized non-zero values?

dominiccooney: Looks like Travis is complaining about cargo fmt.

Yoric: Which values do you have in mind?

Yoric:
(fixing `cargo fmt`)

Issue #423, opened 2019-08-21
Update dependencies


Yoric:
We need to update:

- [lodash](https://github.com/binast/binjs-ref/network/alert/yarn.lock/lodash/open), as there's a security alert (we're probably not vulnerable, but let's err on the safe side);
- [Inflector](https://bugzilla.mozilla.org/show_bug.cgi?id=1575556) for build speed reasons.

PR #433, opened 2019-09-02
Issue #423 - Updating Rust dependencies and releasing new version

Yoric:
A distinct patch will handle the Node dependencies.

### DIFF
--- a/crates/binjs_meta/Cargo.toml
+++ b/crates/binjs_meta/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "binjs_meta"
-version = "0.5.2"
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -133,11 +133,12 @@ color-name@1.1.3:
   integrity sha1-p9BVi9icQveV3UIyj3QIMcpTvCU=
 
 d@1:
```
xtuc: Do you want to change they yarn.lock? It should be in your update for JS deps

Yoric: Oops, that looks like an error, thanks!

PR #434, opened 2019-09-04
Context 0.1: Generating doc from specs

Yoric:


Yoric:
(now with the missing file)

crates/binjs_io/src/context/format.rs
```diff
@@ -0,0 +1,6 @@
+// Fake module. This is only a placeholder for documentation.
+
+// As of this writing, doc inclusion only works on Nightly compiler.
+#![cfg_attr(feature = "unstable", doc(include = "../../../spec/context.md"))]
+
+#![cfg_attr(not(feature = "unstable"), doc = "To generate the format spec documentation, please use `cargo doc --features unstable`. This requires Rust Nightly.")]
```
xtuc: Would be great if we could emit a warning (à la C++), but I believe Rust doesn't support it.

Yoric: A warning if we're generating doc *and* not using Rust nightly?

I don't think it's really necessary. Perhaps we could drop a hyperlink to the online html, though.

PR #435, opened 2019-09-05
bpy encode_dir may now skip errors and copy source files

Yoric:
Use case: building a large corpus of samples for benchmarking purposes from a set of samples that may not be encodable.

### DIFF
--- a/fbssdc/bpy.py
+++ b/fbssdc/bpy.py
@@ -18,8 +18,9 @@ import types
 import argparse
 import json
 import sys
+import shutil
 
-def encode_dir(dict_file, binjs_encode, in_path, out_path):
+def encode_dir(dict_file, binjs_encode, in_path, out_path, skip_errors=True, copy_source=True):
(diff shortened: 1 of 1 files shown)


fbssdc/bpy.py
```diff
@@ -42,13 +43,23 @@ def encode_dir(dict_file, binjs_encode, in_path, out_path):
...
+        format.write(types, string_dict, ty_script, proggy, dest_file)
+
+        # 4. Copy source file
+        if copy_source:
+          shutil.copy(source_path, dest_root)
+      except:
+        if skip_errors:
+          print ('...does not parse')
+        else:
+          pass
```
vdjeric: So skip_errors=True prints "does not parse" but skip_errors=False just causes it to fail silently and continue with the for-loop? Don't you want the operation to stop if skip_errors=False?

Yoric: Oops, you're right.

I meant to rethrow instead of `pass`.

PR #436, opened 2019-09-05
Context 0.1: Getting started on string tables

Yoric:


Yoric:
(now with `cargo fmt`)

crates/binjs_io/src/context/strings.rs
```diff
@@ -0,0 +1,178 @@
...
+                            1 => 1,
+                            _ => return Err(Error::new(ErrorKind::InvalidData, "Invalid escape")),
+                        }
+                    }
+                    byte => byte,
+                };
+                chars.push(byte);
+            }
+            // Attempt to decode string as UTF-8.
+            let string = String::from_utf8(chars)
```
dominiccooney: I don't think ECMAScript makes any guarantee that string data is UTF-8. Should you store this as an array of bytes?

crates/binjs_io/src/context/strings.rs
```diff
@@ -0,0 +1,178 @@
...
+fn test_empty_external_string_dictionary() {
+    let bytes = b"astdict\0";
+    let mut input = std::io::Cursor::new(&bytes);
+    let dictionary = ExternalStringDictionary::read(&mut input).unwrap();
+    assert_eq!(dictionary.strings().len(), 0);
+}
+
+// ----- Writing
+
+impl StringTable {
```
dominiccooney: In practice you might decode many files with one external dictionary. Does it make sense to have a string table point to a couple of tables and handle concatenating them virtually?

Yoric: Yes, that's what I intend to do, but not in this module.

crates/binjs_io/src/context/strings.rs
Yoric: I *think* that string data *in source files* is guaranteed to be representable as UTF-8.

@arai-a, you're the expert, can you confirm/infirm?

arai-a: this basically depends on how the encoder encodes parsed strings.

In term of ECMAScript, it treats the input as a sequence of Unicode code points [1], and there can be lone surrogate `U+D800` - `U+DFFF` that Rust's UTF-8 `String` doesn't accept.
So, lone surrogate should be somehow handled.

Then, in BinAST multipart format we used WTF-8 in the file format, that accepts lone surrogate and encode them in the same way as UTF-8 (so, still, decoding that as UTF-8 fails).
But, in practice, we should validate that it's still in WTF-8 encoding (like, byte sequence `0xff 0x00 0x01` is invalid WTF-8 string), while reading the file.

Then, in current `bpy.py`, we use escaped WTF-8 [2], that I guess used unexpectedly.
So, `\uD800` in JS source is encoded as `\x7f, 'D', '8', '0', '0'` byte sequence in the binjs file.
and all strings there should be valid UTF-8.
but I think we should fix, given escaped WTF-8 isn't supposed to be exposed in external file.

[1] https://tc39.es/ecma262/#sec-source-text
[2] https://github.com/binast/binjs-ref/blob/master/crates/binjs_io/src/escaped_wtf8.rs

Yoric: So, @arai-a, what do you suggest?

Is there we need to fix in `bpy.py` before we proceed? If so, how does Facebook test on their infrastructure that the fix hasn't broken anything?

Otherwise, is this something that we should fix in the Rust implem?

arai-a: if `bpy.py` is kept canonical implementation, we should fix it by either:
  A) throw error/warning on lone surrogate
  B) use WTF-8 (or something?)

for A), testing on Facebook should just check there's no error/warning.
for B), we need to add testcase that string literal contains `\x7F` (escape character in escaped WTF-8, that encoded differently now) and lone surrogates

for Rust implementation, using WTF-8 won't break compatibility
given there's no WTF-8 code, and remaining things are just valid UTF-8,
just that wrongly-encoded lone surrogate and `\x7F` are not supported (this should be valid behavior)

then, inside Rust implementation, we can use escaped WTF-8 here.

Yoric: I vote that we should do A) everywhere for the moment.
@dominiccooney what do you think?

dominiccooney: I don't think @arai-a 's comments about ECMA and Unicode are relevant, because that is about how the parser treats the surface syntax.

I don't know if this speced in BinAST per se, but if you look at what Shift AST which we're built on top of does, it processes escapes before producing strings. For example, see here that there's an escape sequence `\\` in the input but the escape is processed and there is one backslash in the output: https://shift-ast.org/parser.html?parse_type=module&script=%22use%20strict%22%3B%0A%0Alet%20s%20%3D%20%22hello%5C%5Cthere%22%3B%0A

As I understand it an unpaired surrogate like 0xd800 is invalid, so UTF-16 data like 0xd800, 0x0000 is not valid UTF-16. However I can write a JavaScript string literal like '\ud800\u0000' and it means something. You can construct similar examples for UTF-8.

So I think this shows that at the very least you can't expect these strings to be valid ... anything ... and represent what JavaScript string literals can do after they have been parsed.

This should be cleared up. I don't think treating them as UTF-whatever is the answer though.

Yoric: Let me rephrase: I suggested we consider such cases as unspecified-for-the-time-being because that's very much an edge case and we have much more important stuff to work on.

I assumed that parsing as UTF-8 and failing on non-UTF-8 parsable strings was a simple way to do so. Would the pre-treatment by shift prevent this?

dominiccooney: I agree this needs to be straightened out. For now, wouldn't it be simpler to just treat them as bytes?

Yoric: @dominiccooney We can do that here, but we're going to need to convert them to strings somewhere at least for decompression & pretty-printing purposes.

Yoric: Well, moved them to `Vec<u8>`, but we're going to have the same conversation at higher-level whenever I find time to work on encoding/decoding the string prelude!

Yoric:
Thanks for the review!

PR #441, opened 2019-09-11
Getting started on Huffman tables

Yoric:


crates/binjs_io/src/context/huffman.rs
```diff
@@ -0,0 +1,252 @@
...
+            heap.push(Reverse(Node {
+                instances: left.0.instances + right.0.instances,
+                content: NodeContent::Internal(Rc::new((left.0.content, right.0.content))),
+            }));
+        }
+
+        // Convert tree into bit lengths
+        let root = heap.pop().unwrap(); // We have checked above that there is at least one value.
+        let mut bit_lengths = Vec::with_capacity(len);
+        fn aux<T>(bit_lengths: &mut Vec<(T, BitLen)>, depth: u8, node: NodeContent<T>) {
```
dominiccooney: Might be worth putting a FIXME/TODO and an assert here that the length generated is not >20 bits, since this is the earliest you can detect length limits are going to be violated and the code you'll have to change to do limits.

crates/binjs_io/src/context/huffman.rs
```diff
@@ -0,0 +1,252 @@
+#[allow(dead_code)] // Let's get rid of dead code warnings until they make sense.
+use io::statistics::Instances;
+
+use std::cmp::Ordering;
+use std::collections::{BinaryHeap, HashMap};
+use std::hash::Hash;
+use std::rc::Rc;
+
+/// A newtype for `u8` used to count the length of a key in bits.
+#[derive(
```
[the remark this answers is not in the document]
Yoric: Yeah, it's verbose, but I don't know of any such trait.
There used to be something called "int-like", I believe, in another crate, but that crate is now deprecated.

crates/binjs_io/src/context/huffman.rs
```diff
@@ -0,0 +1,252 @@
...
+    content: NodeContent<T>,
+}
+
+/// Contents of a node in the Huffman tree.
+enum NodeContent<T> {
+    /// A value from the stream of values.
+    Leaf(T),
+
+    /// An internal node obtained by joining two subtrees.
+    Internal(Rc<(NodeContent<T>, NodeContent<T>)>), // Invariant: the `Rc` has exactly one owner.
```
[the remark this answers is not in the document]
Yoric: The original reason is that you can un-`Rc` a value without cloning it but you can't un-`Box` it in Rust stable. But I reworked my code and made this irrelevant, so let's get rid of this.

crates/binjs_io/src/context/huffman.rs
Yoric: I thought the idea was to not make the 20 bits part of the spec?

Yoric: Added a mechanism to check that we do not exceed a maximal length.

dominiccooney:
20 bits is part of the spec; how you compute the number 0 ≤ x ≤ 20 is an
implementation detail.

crates/binjs_io/src/context/huffman.rs
```diff
@@ -113,11 +112,20 @@ where
 {
     /// Compute a `Keys` from a sequence of values.
     ///
+    /// Optionally, `max_bit_len` may specify a largest acceptable bit length.
+    /// If `Keys` may not be computed without exceeding this bit length,
+    /// fail with `Err(problemantic_bit_length)`.
```
dominiccooney: Nit: problematic

